PR #4, opened 2017-11-12
Ignore `rustfmt` and all IntelliJ data

sanmai-NL:


djc:
This actually had one too many trailing newlines, so I removed it in 5b180d8.

PR #5, opened 2017-11-12
Add EditorConfig file

sanmai-NL:
This makes any many expectations on file formatting explicit.
Based on the ring library's exemplar.


djc:
Thanks.

PR #6, opened 2017-11-17
Apply strict Clippy and rustc linting

sanmai-NL:
Also resolve lint failures thus found, insofar easily fixed. After resolving the remaining failures we can gate commits on linting success.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -201,6 +201,6 @@ impl Iterator for IdGenerator {
     type Item = RequestId;
     fn next(&mut self) -> Option<Self::Item> {
         self.next += 1;
-        Some(RequestId(format!("A{:04}", self.next % 10000)))
+        Some(RequestId(format!("A{:04}", self.next % 10_000)))
     }
 }
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -36,7 +36,7 @@ impl Client {
     }
 
     pub fn call(self, cmd: Command) -> ResponseStream {
-        let Client { transport, mut state } = self;
+        let Self { transport, mut state } = self;
         let request_id = state.request_ids.next().unwrap();
         let (cmd_bytes, next_state) = cmd.to_parts();
         let future = transport.send(Request(request_id.clone(), cmd_bytes));
@@ -55,8 +55,8 @@ pub struct ResponseStream {
 
 impl ResponseStream {
     pub fn new(future: Send<ImapTransport>, state: ClientState,
-               request_id: RequestId, next_state: Option<State>) -> ResponseStream {
-        ResponseStream {
+               request_id: RequestId, next_state: Option<State>) -> Self {
+        Self {
             future: Some(future),
             transport: None,
             state: Some(state),
@@ -171,8 +171,8 @@ pub struct ClientState {
 }
 
 impl ClientState {
-    pub fn new() -> ClientState {
-        ClientState {
+    pub fn new() -> Self {
+        Self {
             state: State::NotAuthenticated,
             request_ids: IdGenerator::new(),
         }
@@ -188,8 +188,8 @@ pub struct IdGenerator {
 }
 
 impl IdGenerator {
-    pub fn new() -> IdGenerator {
-        IdGenerator { next: 0 }
+    pub fn new() -> Self {
+        Self { next: 0 }
     }
 }
 
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -179,6 +179,10 @@ impl ClientState {
     }
 }
 
+impl Default for ClientState {
+    fn default() -> Self { Self::new() }
+}
+
 pub struct IdGenerator {
     next: u64,
 }
@@ -187,6 +191,10 @@ impl IdGenerator {
     pub fn new() -> IdGenerator {
         IdGenerator { next: 0 }
     }
+}
+
+impl Default for IdGenerator {
+    fn default() -> Self { Self::new() }
 }
 
 impl Iterator for IdGenerator {
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -99,22 +99,18 @@ impl StateStream for ResponseStream {
             let client = Client { transport, state };
             return Ok(Async::Ready(StreamEvent::Done(client)));
         }
-        loop {
-            match transport.poll() {
-                Ok(Async::Ready(Some(rsp))) => {
-                    if let Some(req_id) = rsp.request_id() {
-                        self.done = *req_id == self.request_id;
-                    };
-                    self.transport = Some(transport);
-                    return Ok(Async::Ready(StreamEvent::Next(rsp)));
-                },
-                Ok(Async::Ready(None)) | Ok(Async::NotReady) => {
-                    break;
-                },
-                Err(e) => {
-                    return Err(e);
-                },
-            }
+        match transport.poll() {
+            Ok(Async::Ready(Some(rsp))) => {
+                if let Some(req_id) = rsp.request_id() {
+                    self.done = *req_id == self.request_id;
+                };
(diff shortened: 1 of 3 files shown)


djc:
While I appreciate your efforts to improve the code and make sure the quality level remains high, I don't support this way of going about it. Following all the guidelines out there without exercising judgement is not, in my opinion, a great way to get high quality. For example:

- Replacing type names with `Self` can make code harder to understand, because the actual meaning is hidden behind a layer of indirection.
- Adding trait implementations that are not actually used anywhere means having more code at the expense of having more code (thus making the code base harder to understand).
- I don't like the practice of having `panic!()` used as an expression, since it doesn't actually result anything that matches the stated return type.

On the other hand, I really like the change to use the underscore in the number literal; this actually makes the code easier to read. I can also definitely get behind denying warnings and unused code by default. All the other lint enforcements make me wary of enabling global rules without allowing us to exercise judgement about what's best; we can have a discussion on case-by-case basis, but just pulling everything in because "lints are always a good thing" is not my thing.

sanmai-NL:
Thanks for your view. I did exercise judgment, as I will explain in the following.


## `Self`

No downsides here IMO. Very clear code remains, since the structs required little code in the first place. The downside you mention [wasn't noted in the appropriate place](https://rust-lang-nursery.github.io/rust-clippy/master/index.html#use_self). If you are convinced of that, I think you should propose to add it there. For me, reading `Self` actually is less distracting than repetition. The upside of this change is that a struct name change now requires fewer changes, i.e. better maintainability. Note that you did use `Self` in other places, this change just makes the usage consistent and conventional.

## The `Default` implementations

To clarify, `Default` is only trait you refer to when you mention my implementation traits. These are added as per the Rust API guidelines: [`C-COMMON-TRAITS`](https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#c-common-traits), specifically:
> Note that it is common and expected for types to implement both `Default` and an empty `new` constructor. `new` is the constructor convention in Rust, and users expect it to exist, so if it is reasonable for the basic constructor to take no arguments, then it should, even if it is functionally identical to `default`.

The types are public, so we do not know how they are and aren't used. Wise programmers have discussed the Rust API guidelines at length. Any complaints about it should be directed there. Let's rally around common conventions even if they feel foreign at first.

## `panic!` as expression

This is an important change given the availability of the never type, see [comments on `panic!`](https://doc.rust-lang.org/book/second-edition/ch19-04-advanced-types.html#the-never-type--that-never-returns). It is a better expression of the truth than the unit type here. Similar to how not writing needless `return` statements is idiomatic Rust, it is IMO a good convention to use expressions and have them be of the proper type. The function does not return when panicking, obviously.

## General linting strategy
It is really critical and a maintainability boost in projects to have strict lint checking upfront IMO. Remember that they can always be suppressed via the command line, and first and foremost in specific instances in the code. That would be the better strategy: to suppress lints at an appropriate scope, for example based on your own, non-conventional opinions, and then to motivate that with a comment right there. You will find that in the end, few exceptions are really needed. The people who worked and discussed hard to implement those lint checks are almost always right. If they have made a mistake, then the lint check itself can and should be fixed too.

djc:
I disagree with your premise that it is critical and a maintainability boost in all projects to have strict lint checking. At this stage, there is no stream of contributions that are at risk of destabilizing the code -- you and I are the only contributors so far, with no others in sight. As such, strict lint checking *at this stage* would be making it harder to contribute, not easier.

Thanks for explaining your reasoning about the particular changes. Given these separate explanations, I'm inclined to accept all of these. However, I'd like to ask that you split this up in multiple commits:

* One commit for the change in the number literal
* One commit for `Self` renaming, explaining the rationale in the commit message
* One commit for `Default` trait implementations, plus rationale
* One commit for removing the semi-colon from `panic!()`, plus rationale
* One commit adding all the lint `deny`s you have

We can add `warn(missing_docs)` (or `deny(missing_docs)`) after the documentation is filled in. For now I don't think I want to add any crate-wide clippy-related attributes.

sanmai-NL:
You're right, it's not always critical. It is in larger projects and I do take for granted this project will turn into one. We'll have to see. One side note: making contributions harder, at first, can be a good thing (like other contributing guidelines).

You're also right about the need to split the commit.

src/lib.rs
```diff
@@ -1,3 +1,13 @@
+#![deny(warnings)]
+#![deny(unused)]
+#![deny(future_incompatible)]
+#![deny(bad_style)]
+#![warn(missing_docs)]
```
djc: As mentioned in a comment, I want to refrain from including `warn(missing_docs)` as well as the others below that for now. (I'm not sure where `box_pointers` is coming from, or why we would need it.)

sanmai-NL: Sorry, will fix that.

sanmai-NL: `box_pointers` is a lint that prevents use of owned heap memory. This restriction seems unneeded for this library, but it exists as a lint. Due to a bug in `rustc` it is not yet included in `deny(warnings)`. I've removed the exception for now.

sanmai-NL: Do you understand that adding the crate-wide Clippy attributes does no harm? It requires no extra dependencies, for example. The attributes are only effective when one installs `cargo-clippy` as a standalone tool and runs it on the project. Since there is no CI yet, the attributes have no effect on you should you choose to write new code that Clippy does not like. By adding those attributes you do give guidance when Clippy is used by a contributor in the future. Not having the attributes means that as I write new code, I'd have to put in more effort to lint it before filing a PR. Using Clippy in this way, with attributes in the code, is the best practice. Could you reconsider this please?

src/client/mod.rs
```diff
@@ -189,6 +193,10 @@ impl IdGenerator {
     }
 }
 
+impl Default for IdGenerator {
```
djc, line 192: Oh, BTW, looks like Clippy actually suggests having `Default` derived for this. Any particular reason you didn't do so here?

src/lib.rs
djc: As far as I can tell, the only effect of these attributes is whether Clippy treats things it finds as warning or error. I don't really understand how that would impede your work -- you can lint them just as warnings, I'd say. Certainly the way I run Clippy I prefer to get warnings rather than errors.

src/lib.rs
```diff
@@ -2,7 +2,6 @@
 #![deny(unused)]
 #![deny(future_incompatible)]
 #![deny(bad_style)]
-#![warn(missing_docs)]
```
djc: Please squash this into your "general linting strategy" commit.

src/lib.rs
```diff
@@ -2,7 +2,6 @@
 #![deny(unused)]
 #![deny(future_incompatible)]
 #![deny(bad_style)]
-#![allow(box_pointers)]
```
djc: Please squash this into your "general linting strategy" commit.

src/client/mod.rs
sanmai-NL: AFAIK deriving `Default` means
```rust
impl Default for Struct {
    fn default() -> Self {
        Self {
            field_1: FieldType1::default(), 
            // and so on
        }
    }
}
```
It doesn't call the constructor, since that concept is sadly only conventional. It appears we can depend on that derived `default` methods would return the same values as your current constructors though. In that case, we can redefine your constructors in terms of the `default` methods.

djc: Ah, makes sense. Yes, I think it would be nice if we `derive(Default)` and then call `Self::default()` in `new()` (effectively reversing the dependency).

sanmai-NL: @djc: You can merge the existing branch, then I'll make the changes to `imap-proto` required for deriving `Default` in this crate.

djc:
I've taken most of this and rebased it onto master. Thanks!

PR #12, opened 2018-02-07
[WIP] Add basic example

sanmai-NL:
There are two `todo`s in the code. Can you take a look and advise?

### DIFF
--- a/tokio-imap/examples/basic.rs
+++ b/tokio-imap/examples/basic.rs
@@ -1,0 +1,136 @@
+extern crate futures;
+extern crate futures_state_stream;
+extern crate tokio;
+extern crate tokio_imap;
+
+use futures::future::Future;
+use futures_state_stream::StateStream;
+use std::error::Error;
+use std::fmt::{self, Display, Formatter};
+use std::io;
+use tokio::executor::current_thread;
+use tokio_imap::ImapClient;
+use tokio_imap::client::builder::{CommandBuilder, FetchBuilderAttributes, FetchBuilderMessages,
+                                  FetchBuilderModifiers};
+use tokio_imap::client::connect;
+use tokio_imap::proto::ResponseData;
+use tokio_imap::types::{Attribute, AttributeValue, Response};
+
+fn process_email(responsedata: &ResponseData) {
+    if let Response::Fetch(_, ref attr_vals) = *responsedata.parsed() {
+        for val in attr_vals.iter() {
+            match *val {
+                AttributeValue::Uid(u) => {
+                    eprintln!("E-mail UID: {}", u);
+                },
+                AttributeValue::Rfc822(Some(src)) => {
+                    eprintln!("E-mail body length: {}", src.to_vec().len());
+                },
+                _ => (),
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum ImapError {
+    Connect { cause: io::Error },
+    Login { cause: io::Error },
+    Select { cause: io::Error },
+    UidFetch { cause: io::Error },
+}
+
+impl Error for ImapError {
+    fn description(&self) -> &'static str {
+        ""
+    }
+
+    fn cause(&self) -> Option<&Error> {
+        match *self {
+            ImapError::Connect { ref cause }
+            | ImapError::Login { ref cause }
+            | ImapError::Select { ref cause }
+            | ImapError::UidFetch { ref cause } => Some(cause),
+        }
+    }
+}
+
+impl Display for ImapError {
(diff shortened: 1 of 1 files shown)


examples/basic.rs
```diff
@@ -0,0 +1,128 @@
...
+                    eprintln!("E-mail body length: {}", src.to_vec().len());
+                }
+                _ => (),
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum SyncError {
```
djc: Since all of these errors are of the same time, I don't think having the `SyncError` type in this example makes a lot of sense.

examples/basic.rs
```diff
@@ -0,0 +1,128 @@
...
+    };
+    eprintln!("Will connect to {}", server);
+    eprintln!("Will login as {}", login);
+    let mut core = Core::new().expect("Failed to create a core. ");
+    let handle = core.handle();
+    let future_imap = Client::connect(server, &handle)
+        .map_err(|cause| SyncError::Connect { cause })
+        .and_then(move |(client, response)| {
+            eprintln!("Connection attempt response: {:#?}", response);
+            // todo What is the correct order of the collect and map_err calls?
```
djc: If you just remove the whole deal with `SyncError` (which isn't really relevant for tokio-imap usage), you shouldn't even need the `map_err()`, solving this problem.

examples/basic.rs
```diff
@@ -0,0 +1,128 @@
...
+            // todo Why does client.call() not return an Err when it should, continuing the chain?
```
djc: I'm not sure what you mean by "when it should". The return type for `Client::call()` is just `StateStream<Item = ResponseData, State = Client, Error = io::Error>`; it can yield errors when polled, but, since all of the actual action is deferred to when the `StateStream` is polled, there will be no errors on just calling `Client::call()` itself.

examples/basic.rs
sanmai-NL: Of the same type you mean? The variants allow pinpointing what error occurred specifically.

examples/basic.rs
sanmai-NL: How can I stop the futures chain when e.g. the login command failed?

sanmai-NL:
Yes it is, suddenly the project didn’t compile anymore since the glob imports change changed the API. That must’ve been inadvertent.

examples/basic.rs
djc: Have you read the documentation on what `and_then()` means? https://docs.rs/futures/0.1.16/futures/future/trait.Future.html#method.and_then

examples/basic.rs
djc: Okay, I guess that makes sense. In this case it would be more idiomatic to have a tuple variants rather than struct variants, though.

sanmai-NL:
@djc: I’ve updated the example code. What I still want to add is error handling based on individual responses. This is particularly helpful for an example. I couldn’t get `map_err` to work on the futures chain though. What solutions do you see?

djc:
Can you rebase this on top of current master? Then I can take a look.

sanmai-NL:
@djc: done.

You can run the basic example on Unix with with:

```sh
 printf "%s\n" "server" "login" "password" "Inbox" | \
      time env RUST_BACKTRACE=1 cargo run --release --package tokio-imap --example basic | less
```

djc:
I'm not sure what you mean by the error handling question, exactly. Can you give an example?

There's a bunch of cosmetic stuff that I'd want to change. Do you want me to write it up or shall I just edit out after you're done? I think this is probably already quite useful to have as a minimal example, even without the added error handling -- no sense to necessarily block the PR on that.

sanmai-NL:
I want be able to signal every possible error state in my error enum. Right now, the chain stops if at some point an unexpected IMAP response happens, but there is no user-friendly way to explain what went wrong. 

I have full confidence in you to adapt the example reasonably, please do. While it is not a blocker for the example, the example is a starting point for me to start integrating your code in our own software. For that, I must be able to signal various error states to the user. I can also image many other users of this library will be interested in more fine-grained error handling.

djc:
Sorry, I still cannot wrap my head around exactly what you want with the error handling. Maybe write up an issue describing what you're looking for with code examples or example stdout output?

sanmai-NL:
@djc: I want the person who runs the example to able to answer the question: did login fail or did the request mailbox not exist? I want specific error conditions, not cryptic IMAP responses printed to the console.

Maybe you can just merge this and I’ll revisit the error handling wishes.

djc:
I rebased your commits as 4119717f9256b5d942adefb6ed99627ddda124a8, and pushed them to master. Thanks!

PR #14, opened 2018-02-09
Tokio upgrade, error handling, Client API refactor

djc:
@jonhoo, @sanmai-NL, others: please review.

- Upgrades from tokio-core to the new tokio reform
- Handle connecting without string formatting the server address
- Improves error handling to have fewer `unwrap()` calls (and annotate the safe ones)
- Decouple `ResponseStream` from `Client` type
- Rename `Client` to `TlsClient`, move behavior into a trait called `ImapClient`

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -95,8 +95,8 @@ impl StateStream for ResponseStream {
         };
         if self.done {
             let mut state = self.state.take().unwrap();
-            if self.next_state.is_some() {
-                state.state = self.next_state.take().unwrap();
+            if let Some(next_state) = self.next_state.take() {
+                state.state = next_state;
             }
             let client = Client { transport, state };
             return Ok(Async::Ready(StreamEvent::Done(client)));
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -8,9 +8,8 @@ use native_tls::TlsConnector;
 use std::io;
 use std::net::ToSocketAddrs;
 
-use tokio_core::net::{TcpStream, TcpStreamNew};
+use tokio::net::{ConnectFuture, TcpStream};
 use tokio_io::AsyncRead;
-use tokio_core::reactor::Handle;
 use tokio_tls::{ConnectAsync, TlsConnectorExt};
 
 use imap_proto::{Request, RequestId, State};
@@ -31,11 +30,10 @@ pub struct Client {
 }
 
 impl Client {
-    pub fn connect(server: &str, handle: &Handle) -> ConnectFuture {
+    pub fn connect(server: &str) -> ImapConnectFuture {
         let addr = format!("{}:993", server);
         let addr = addr.to_socket_addrs().unwrap().next().unwrap();
-        let stream = TcpStream::connect(&addr, handle);
-        ConnectFuture::TcpConnecting(stream, server.to_string())
+        ImapConnectFuture::TcpConnecting(TcpStream::connect(&addr), server.to_string())
     }
 
     pub fn call(self, cmd: Command) -> ResponseStream {
@@ -120,36 +118,36 @@ impl StateStream for ResponseStream {
     }
 }
 
-pub enum ConnectFuture {
-    #[doc(hidden)] TcpConnecting(TcpStreamNew, String),
+pub enum ImapConnectFuture {
+    #[doc(hidden)] TcpConnecting(ConnectFuture, String),
     #[doc(hidden)] TlsHandshake(ConnectAsync<TcpStream>),
     #[doc(hidden)] ServerGreeting(Option<ImapTransport>),
 }
 
-impl Future for ConnectFuture {
+impl Future for ImapConnectFuture {
     type Item = (Client, ResponseData);
     type Error = io::Error;
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         let mut new = None;
-        if let ConnectFuture::TcpConnecting(ref mut future, ref domain) = *self {
+        if let ImapConnectFuture::TcpConnecting(ref mut future, ref domain) = *self {
             let stream = try_ready!(future.poll());
             let ctx = TlsConnector::builder().unwrap().build().unwrap();
             let future = ctx.connect_async(domain, stream);
-            new = Some(ConnectFuture::TlsHandshake(future));
+            new = Some(ImapConnectFuture::TlsHandshake(future));
         }
         if new.is_some() {
             *self = new.take().unwrap();
         }
-        if let ConnectFuture::TlsHandshake(ref mut future) = *self {
+        if let ImapConnectFuture::TlsHandshake(ref mut future) = *self {
             let transport = try_ready!(future.map_err(|e| {
                 io::Error::new(io::ErrorKind::Other, e)
             }).poll()).framed(ImapCodec::default());
-            new = Some(ConnectFuture::ServerGreeting(Some(transport)));
+            new = Some(ImapConnectFuture::ServerGreeting(Some(transport)));
         }
         if new.is_some() {
             *self = new.take().unwrap();
         }
-        if let ConnectFuture::ServerGreeting(ref mut wrapped) = *self {
+        if let ImapConnectFuture::ServerGreeting(ref mut wrapped) = *self {
             let msg = try_ready!(wrapped.as_mut().unwrap().poll()).unwrap();
             return Ok(Async::Ready((Client {
                 transport: wrapped.take().unwrap(),
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -39,7 +39,7 @@ impl Client {
 
     pub fn call(self, cmd: Command) -> ResponseStream {
         let Self { transport, mut state } = self;
-        let request_id = state.request_ids.next().unwrap();
+        let request_id = state.request_ids.next().unwrap(); // safe: never returns Err
         let (cmd_bytes, next_state) = cmd.into_parts();
         let future = transport.send(Request(request_id.clone(), cmd_bytes));
         ResponseStream::new(future, state, request_id, next_state)
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -14,7 +14,7 @@ use tokio_tls::{ConnectAsync, TlsConnectorExt};
 
 use imap_proto::{Request, RequestId, State};
 use imap_proto::builders::command::Command;
-use proto::{ImapCodec, ImapTls, ResponseData};
+use proto::{ImapCodec, ImapTls, ImapTransport, ResponseData};
 
 pub mod builder {
     pub use imap_proto::builders::command::{CommandBuilder, FetchBuilderAttributes,
@@ -24,21 +24,20 @@ pub mod builder {
 }
 
 
-pub struct Client {
-    transport: ImapTls,
-    state: ClientState,
+pub fn connect(server: &str) -> io::Result<ImapConnectFuture> {
+    let addr = (server, 993).to_socket_addrs()?.next().ok_or_else(|| {
+        io::Error::new(io::ErrorKind::Other, format!("no IP addresses found for {}", server))
+    })?;
+    Ok(ImapConnectFuture::TcpConnecting(TcpStream::connect(&addr), server.to_string()))
 }
 
-impl Client {
-    pub fn connect(server: &str) -> io::Result<ImapConnectFuture> {
-        let addr = (server, 993).to_socket_addrs()?.next().ok_or_else(|| {
-            io::Error::new(io::ErrorKind::Other, format!("no IP addresses found for {}", server))
-        })?;
-        Ok(ImapConnectFuture::TcpConnecting(TcpStream::connect(&addr), server.to_string()))
-    }
+pub trait ImapClient {
+    type Transport: ImapTransport;
+    fn into_parts(self) -> (Self::Transport, ClientState);
+    fn rebuild(transport: Self::Transport, state: ClientState) -> Self;
 
-    pub fn call(self, cmd: Command) -> ResponseStream {
-        let Self { transport, mut state } = self;
+    fn call(self, cmd: Command) -> ResponseStream<Self> where Self: ImapClient + Sized {
+        let (transport, mut state) = self.into_parts();
         let request_id = state.request_ids.next().unwrap(); // safe: never returns Err
         let (cmd_bytes, next_state) = cmd.into_parts();
         let future = transport.send(Request(request_id.clone(), cmd_bytes));
@@ -46,17 +45,35 @@ impl Client {
     }
 }
 
-pub struct ResponseStream {
-    future: Option<Send<ImapTls>>,
-    transport: Option<ImapTls>,
+pub struct Client {
+    transport: ImapTls,
+    state: ClientState,
+}
+
+impl ImapClient for Client {
+    type Transport = ImapTls;
+
+    fn into_parts(self) -> (ImapTls, ClientState) {
+        let Self { transport, state } = self;
+        (transport, state)
+    }
+
+    fn rebuild(transport: ImapTls, state: ClientState) -> Client {
+        Client { transport, state }
+    }
+}
+
+pub struct ResponseStream<E> where E: ImapClient {
+    future: Option<Send<E::Transport>>,
+    transport: Option<E::Transport>,
     state: Option<ClientState>,
     request_id: RequestId,
     next_state: Option<State>,
     done: bool,
 }
 
-impl ResponseStream {
-    pub fn new(future: Send<ImapTls>, state: ClientState,
+impl<E> ResponseStream<E> where E: ImapClient {
+    pub fn new(future: Send<E::Transport>, state: ClientState,
                request_id: RequestId, next_state: Option<State>) -> Self {
         Self {
             future: Some(future),
@@ -69,9 +86,9 @@ impl ResponseStream {
     }
 }
 
-impl StateStream for ResponseStream {
+impl<E> StateStream for ResponseStream<E> where E: ImapClient {
     type Item = ResponseData;
-    type State = Client;
+    type State = E;
     type Error = io::Error;
     fn poll(&mut self) -> Poll<StreamEvent<Self::Item, Self::State>, Self::Error> {
         match self.future.take() {
@@ -98,8 +115,7 @@ impl StateStream for ResponseStream {
             if let Some(next_state) = self.next_state.take() {
                 state.state = next_state;
             }
-            let client = Client { transport, state };
-            return Ok(Async::Ready(StreamEvent::Done(client)));
+            return Ok(Async::Ready(StreamEvent::Done(E::rebuild(transport, state))));
         }
         match transport.poll() {
             Ok(Async::Ready(Some(rsp))) => {
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -30,9 +30,11 @@ pub struct Client {
 }
 
 impl Client {
-    pub fn connect(server: &str) -> ImapConnectFuture {
-        let addr = (server, 993).to_socket_addrs().unwrap().next().unwrap();
-        ImapConnectFuture::TcpConnecting(TcpStream::connect(&addr), server.to_string())
+    pub fn connect(server: &str) -> io::Result<ImapConnectFuture> {
+        let addr = (server, 993).to_socket_addrs()?.next().ok_or_else(|| {
+            io::Error::new(io::ErrorKind::Other, format!("no IP addresses found for {}", server))
+        })?;
+        Ok(ImapConnectFuture::TcpConnecting(TcpStream::connect(&addr), server.to_string()))
     }
 
     pub fn call(self, cmd: Command) -> ResponseStream {
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -94,7 +94,7 @@ impl StateStream for ResponseStream {
             Some(mut transport) => transport,
         };
         if self.done {
-            let mut state = self.state.take().unwrap();
+            let mut state = self.state.take().unwrap(); // safe: initialized from start
             if let Some(next_state) = self.next_state.take() {
                 state.state = next_state;
             }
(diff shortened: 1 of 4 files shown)


src/client/mod.rs
```diff
@@ -25,39 +24,58 @@ pub mod builder {
 }
 
 
-pub struct Client {
-    transport: ImapTransport,
-    state: ClientState,
+pub fn connect(server: &str) -> io::Result<ImapConnectFuture> {
+    let addr = (server, 993).to_socket_addrs()?.next().ok_or_else(|| {
+        io::Error::new(io::ErrorKind::Other, format!("no IP addresses found for {}", server))
```
jonhoo, line 29: I think `io::ErrorKind::AddrNotAvailable` is preferred here.

src/client/mod.rs
```diff
@@ -25,39 +24,58 @@ pub mod builder {
...
 
-impl Client {
-    pub fn connect(server: &str, handle: &Handle) -> ConnectFuture {
-        let addr = format!("{}:993", server);
-        let addr = addr.to_socket_addrs().unwrap().next().unwrap();
-        let stream = TcpStream::connect(&addr, handle);
-        ConnectFuture::TcpConnecting(stream, server.to_string())
-    }
+pub trait ImapClient {
+    type Original;
```
jonhoo: What is the purpose of `Original` here? Why can't you just use `Self`?

src/proto.rs
```diff
@@ -91,3 +93,11 @@ impl ResponseData {
...
     }
 }
+
+pub type ImapTls = Framed<TlsStream<TcpStream>, ImapCodec>;
+
+impl ImapTransport for ImapTls {}
+
+pub trait ImapTransport: futures::Stream<Item = ResponseData, Error = io::Error> +
+                         futures::Sink<SinkItem = Request, SinkError = io::Error> {
+}
```
jonhoo, line 103: If only we had trait aliases :(

src/client/mod.rs
```diff
@@ -25,39 +24,58 @@ pub mod builder {
...
 
-pub struct ResponseStream {
-    future: Option<Send<ImapTransport>>,
-    transport: Option<ImapTransport>,
+pub struct TlsClient {
+    transport: ImapTls,
+    state: ClientState,
+}
+
+impl ImapClient for TlsClient {
```
jonhoo, line 53: So there is (currently) no `ImapClient` that isn't `TlsClient`? Are you expecting there to be? If not, why the trait?

src/client/mod.rs
```diff
@@ -70,14 +88,13 @@ impl ResponseStream {
     }
 }
 
-impl StateStream for ResponseStream {
+impl<E> StateStream for ResponseStream<E> where E: ImapClient {
     type Item = ResponseData;
-    type State = Client;
+    type State = E::Original;
```
jonhoo: If you got rid of `Original`, this would just be `E`.

sanmai-NL:
Kudos for this work. I’ve scheduled my review tomorrow 16:00. Assuming you’ve addressed Jon’s comments by then. If you wish me to review at a later time, don’t hesitate to tell.

src/client/mod.rs
djc: Hmm. I'm not entirely convinced. The [documentation](https://doc.rust-lang.org/std/io/enum.ErrorKind.html) there is "A nonexistent interface was requested or the requested address was not local"; this is about finding an address for a host (not an interface), and it has nothing to do with whether the address is local or not.

src/client/mod.rs
djc: Ah, that shows I still don't fully grok the trait system. Nice catch, I'll fix it up.

src/proto.rs
djc: Yeah, tried that first. Oh, well, at least this works...

src/client/mod.rs
djc: Yup, this is gone now.

src/client/mod.rs
djc: So my goal was to untie `ResponseStream` from the fixed `Client` type. `Client` as is is restricted to exactly `ImapTls`, which seems a bit limiting when the real constraint is just `ImapTransport`. I have two use cases in mind: some kind of mock implementation of `ImapTransport` that can help with testing, and an implementation that can deal with non-TLS TCP streams. Also, the way tokio works really necessitated separating the state from the transport because you have to "give away" the transport for a bit while sending; so, another use case might be if you wanted to implement some kind of alternative client where you have more state to keep around. (I'll add some of this to the commit message.)

The alternative approach is here to just parametrize the `Client` type with `T: ImapTransport`, which I implemented and updated mailsync for, but I found that having to parametrize with `ImapTls` is unergonomic boilerplate for what really seems like the most common case.

Then, there is the issue of naming `Client`; given the above, I was looking for clear names for both of the trait and the default `ImapTls` implementation, and this was the best I could come up with. Better suggestions welcome!

src/client/mod.rs
```diff
@@ -88,19 +105,19 @@ impl StateStream for ResponseStream {
...
+        let mut transport = match self.transport.take() {
+            None => return Ok(Async::NotReady),
+            Some(mut transport) => transport,
+        };
         if self.done {
-            let mut state = self.state.take().unwrap();
+            let mut state = self.state.take().unwrap(); // safe: initialized from start
             if self.next_state.is_some() {
-                state.state = self.next_state.take().unwrap();
+                state.state = self.next_state.take().unwrap(); // safe: checked on previous line
```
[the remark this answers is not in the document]
djc: Much better, not sure why I couldn't come up with that! I'll fix it up.

djc:
@jonhoo thanks for the thoughtful comments so far! @sanmai-NL tomorrow at 16:00 should be good, thanks for making time for this.

src/client/mod.rs
jonhoo: Ah. yes, sorry, for some reason I was thinking of this as binding. Never mind.

src/client/mod.rs
```diff
@@ -24,39 +24,56 @@ pub mod builder {
 }
 
 
-pub struct Client {
-    transport: ImapTls,
-    state: ClientState,
+pub fn connect(server: &str) -> io::Result<ImapConnectFuture> {
```
djc: BTW, curious for any thoughts on this: `connect()` as a free function in the `client` module that returns a `TlsClient`, or `connect()` as a static method on the `TlsClient` itself?

src/client/mod.rs
jonhoo: If you don't expect there to ever be many implementations of `ImapClient`, only of `ImapTransport`, I'd say the "right" pattern is to have `ImapClient` be a struct parameterized by `T: ImapTransport` (as you outline above). It's true that this puts some burden on users of the library though. There are two workarounds for that:

 - Provide a newtype wrapper. For example:
   
   ```rust
   pub type TlsClient = ImapClient<TlsTransport>;
   ```
 - Store the transport as a `Box<ImapTransport>` inside the `ImapClient`. That way, `ImapClient` does not need to be generic, and the (marginal) cost of dynamic dispatch for network calls.

src/client/mod.rs
jonhoo: I prefer the latter, because then I can write:
```rust
use tokio_imap::TlsClient;
// ...
    let c = TlsClient::connect(...);
// ...
fn foo(c: TlsClient) { ... }
```
instead of
```rust
use tokio_imap::{self, TlsClient};
// ...
    let c = tokio_imap::connect(...);
// ...
fn foo(c: TlsClient) { ... }
```
but it's a pretty minor preference.

src/client/mod.rs
djc: Why would it be better, in your opinion, to have `ImapClient` be a parameterized struct over a trait? This at least doesn't support the third use case I mentioned, and I'm not seeing that many downsides from the current arrangement.

src/client/mod.rs
djc: Yeah, I think I'm back to leaning that way, too.

src/client/mod.rs
jonhoo: I guess to me the biggest downside of a trait here is that I now can't have, say, a struct that holds an `ImapClient`. Instead, it'll either have to hold a `TlsClient`, a `Box<ImapClient>`, or be generic over `T: ImapClient`, all of which seem a little weird. Instead, holding an `ImapClient` (or potentially an `ImapClient<T>`, though I don't think that's necessary) feels more natural somehow.

djc: Yeah, and so I'd argue, given what I said above, that that is on purpose, and this design should point you in the direction of instead `impl ImapClient for YourClient` -- but maybe that is too onerous/unidiomatic/weird. Conceptually, given my experience with the tokio stack, it feels like the right design. Can you comment on how that would fit your use cases?

djc: (Actually, that might not fit with what I have now -- before, `ClientState` was abstracted out as an associated type `ImapClient::State`, which I think would make this more interesting. Would need some API to get a `&mut ClientState`, I guess.)

jonhoo: I'm confused -- surely it would be very uncommon for users of `tokio-imap` to ever `impl ImapClient` for their own things anywhere? They're likely to write programs that *use* an IMAP connection, probably in a binary, but not likely to write their own `ImapClient`?

I think of traits as behaviors. `ImapClient` as a trait means "things that can behave as an `ImapClient`. It's true that `TlsImapClient` can behave as an `ImapClient`, but it does so in exactly the same way that a `MockImapClient` behaves like one (i.e., none of them overwrite `call`). The *real* difference is just in their transport. Because of that I'd argue `ImapClient<T: ImapTransport>` makes more sense. Now, you have a single, shared implementation of an IMAP client that can operate on top of anything that can take IMAP commands and return IMAP responses.

As I argued above, you could then even get rid of the `<T>` by having `ImapClient` store a `Box<ImapTransport>` internally.

jonhoo: I'm not sure how `ClientState` factors into this. Do you expect users of the library to ever fiddle with `ClientState` directly?

djc: @jonhoo so note that `ImapClient::call()` takes ownership of self (and goes on to destroy it). To me, that makes just putting `TlsClient` in your structs pretty painful. Best case, you have to wrap it in an `Option` and work with `take()` to take it out and put it in again. `impl ImapClient` is one way to make this more explicit and maybe slightly less painful.

Users of the library might want to inspect the `ClientState.state`, but otherwise they don't have to fiddle with it. However, `ImapClient::call()` and `ResponseStream` need access to it.

jonhoo: Ah, yes, the terrible terrible "move out of `self`" part of futures. Yeah, I guess maybe you're right that having the user write `impl ImapClient` would make that easier. It's just really weird to me, because their wrapper type that *uses* and `ImapClient` isn't itself an `ImapClient`. Not really. I think I'd actually argue that the right thing to do is to let the user *choose* how they want to use the `ImapClient` you give them, with the knowledge that it consumes and returns `self`.

The internals of `ImapClient::call` and `ResponseStream` are both hidden from the users, so it's fine for them to touch `ClientState`. I was more trying to understand your point about `ClientState` in https://github.com/djc/tokio-imap/pull/14#discussion_r167351289.

src/client/mod.rs
```diff
@@ -120,36 +118,36 @@ impl StateStream for ResponseStream {
...
+    #[doc(hidden)] TcpConnecting(ConnectFuture, String),
     #[doc(hidden)] TlsHandshake(ConnectAsync<TcpStream>),
     #[doc(hidden)] ServerGreeting(Option<ImapTransport>),
 }
 
-impl Future for ConnectFuture {
+impl Future for ImapConnectFuture {
     type Item = (Client, ResponseData);
     type Error = io::Error;
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
```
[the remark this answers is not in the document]
sanmai-NL: I spent some time to optimize this `poll` implementation, and came up with this:

```rust
impl Future for ImapConnectFuture {
    type Item = (TlsClient, ResponseData);
    type Error = io::Error;
    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
        loop {
            break match *self {
                ref mut future_now @ ImapConnectFuture::TlsHandshake(_) => {
                    let transport = if let &mut ImapConnectFuture::TlsHandshake(ref mut future) = future_now {
                        let mut result_future = future.map_err(|e| {
                            io::Error::new(io::ErrorKind::Other, e)
                        });
                        try_ready!(result_future.poll()).framed(ImapCodec::default())
                    } else { unreachable!() };
                    *future_now = ImapConnectFuture::ServerGreeting(Some(transport));
                    continue;
                },
                ref mut future_now @ ImapConnectFuture::TcpConnecting(_, _) => {
                    let future_next = if let &mut ImapConnectFuture::TcpConnecting(ref mut future, ref domain) = future_now {
                        let stream = try_ready!(future.poll());
                        let ctx = TlsConnector::builder().unwrap().build().unwrap();
                        ctx.connect_async(domain.as_str(), stream)
                    } else { unreachable!() };
                    *future_now = ImapConnectFuture::TlsHandshake(future_next);
                    continue;
                },
                ImapConnectFuture::ServerGreeting(ref mut wrapped) => {
                    let msg = try_ready!(wrapped.as_mut().unwrap().poll()).unwrap();
                    Ok(Async::Ready((TlsClient {
                        transport: wrapped.take().unwrap(),
                        state: ClientState::new(),
                    }, msg)))
                }
            };
        }
    }
}
```
It removes the `Option` layer along with some unwraps and duplication, as well as the extra mutable variable binding `new`.

src/client/mod.rs
```diff
@@ -31,8 +31,7 @@ pub struct Client {
 
 impl Client {
     pub fn connect(server: &str) -> ImapConnectFuture {
-        let addr = format!("{}:993", server);
-        let addr = addr.to_socket_addrs().unwrap().next().unwrap();
+        let addr = (server, 993).to_socket_addrs().unwrap().next().unwrap();
```
sanmai-NL: This will panic depending on the value of `server`. `server` can be untrusted input. It in is our software: users enter an IMAP server address and it could be anything. So please return a `Result` instead.

src/proto.rs
```diff
@@ -48,7 +48,10 @@ impl<'a> Decoder for ImapCodec {
             IResult::Incomplete(_) => {
                 return Ok(None);
             },
-            IResult::Error(err) => panic!("error {} during parsing of {:?}", err, buf),
+            IResult::Error(err) => {
+                return Err(io::Error::new(io::ErrorKind::Other,
+                                          format!("error {} during parsing of {:?}", err, buf)));
```
sanmai-NL, line 53: `"error {}` -> `"{}` (DRY). Also, I prefer to have precise capitalization, punctuation etc. in any text, especially if users may see it. Admittedly, even this wasn’t even done in `std`, which has poorly written error messages like (`os error 14`) all over the place.

src/client/mod.rs
```diff
@@ -74,9 +74,8 @@ impl StateStream for ResponseStream {
     type State = Client;
     type Error = io::Error;
     fn poll(&mut self) -> Poll<StreamEvent<Self::Item, Self::State>, Self::Error> {
-        if self.future.is_some() {
-            let mut future = self.future.take().unwrap();
-            match future.poll() {
+        match self.future.take() {
+            Some(mut future) => match future.poll() {
```
sanmai-NL, line 95: `self.future.map(|future| { ... })` would be even more concise, wouldn’t it?

src/proto.rs
sanmai-NL: Interesting remark Jon. We do not have [trait aliases](https://github.com/rust-lang/rfcs/blob/master/text/1733-trait-alias.md) yet, but I concur that this works too. I don’t see trait aliases being a boon here but perhaps I’m missing something.

src/client/mod.rs
sanmai-NL: I tend to favor Jon’s alternative `ImapClient<T>` in principle/i.e. because of the intuitiveness of it. Dirkjan, you explain the `call` method consumes `self`, but does it need to? Can’t it e.g. call `rebuild` or return both the `ResponseStream` and e.g. `self.transport`?

More generally, we can simply stick with Dirkjans current design instead of ruminating it, since Tokio is in flux anyway and `tokio-io` will be history.

src/client/mod.rs
```diff
@@ -120,36 +118,36 @@ impl StateStream for ResponseStream {
     }
 }
 
-pub enum ConnectFuture {
-    #[doc(hidden)] TcpConnecting(TcpStreamNew, String),
+pub enum ImapConnectFuture {
+    #[doc(hidden)] TcpConnecting(ConnectFuture, String),
```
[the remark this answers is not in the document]
djc: The enum variants here aren't part of any public API, they're just states in the state machine for this future. As such, I don't think it makes sense to document them.

src/client/mod.rs
```diff
@@ -120,36 +118,36 @@ impl StateStream for ResponseStream {
...
+impl Future for ImapConnectFuture {
     type Item = (Client, ResponseData);
     type Error = io::Error;
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         let mut new = None;
-        if let ConnectFuture::TcpConnecting(ref mut future, ref domain) = *self {
+        if let ImapConnectFuture::TcpConnecting(ref mut future, ref domain) = *self {
             let stream = try_ready!(future.poll());
             let ctx = TlsConnector::builder().unwrap().build().unwrap();
             let future = ctx.connect_async(domain, stream);
```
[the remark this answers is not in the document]
djc: I'd prefer to keep this PR focused on the changes made by this PR. If you have other suggestions, please file them as separate issues or PRs.

djc:
@sanmai-NL thanks for the review. Some of your comments aren't connected to the changes they're commenting on (and I can't reply to them in situ) -- it looks like you conducted multiple reviews and globally-requested changes each time. Instead, please comment on the areas of the code specifically, so we can have a discussion. Also, please focus on the changes proposed in this PR -- other suggestions are obviously welcome, but are IMO a distraction from getting this PR merged.

src/client/mod.rs
sanmai-NL: Also favor the second alternative. Even better (for debugging and forward compatibility) would be for `connect` to use a `server` field carried by `TlsClient` rather than a formal parameter.

src/client/mod.rs
djc: Maybe -- it's not a great fit conceptually, though, since fundamentally we're taking out the future here rather than converting it to an `Option` of a different type.

sanmai-NL:
I reviewed each commit and each comment pertained to a specific line. I had no idea GitHub would make the comments global. Weird, since the review/comment box is only shown when reviewing specific commits.

Not sure what things I commented on that do not appear in this PR?

src/client/mod.rs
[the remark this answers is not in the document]
djc: Please file this as a separate PR, so we can debate it separately. I think I'm going to be somewhat insensitive to performance optimizations of this type unless you've done the legwork of benchmarking, concluding that this is a hotspot, and that these changes will meaningfully improve performance. As it is, I'd rather optimize for readability (although not at the expense of everything else).

src/proto.rs
djc: I can get behind the removal of "error".

src/client/mod.rs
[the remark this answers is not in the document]
sanmai-NL: They could be documented regardless, for contributors as well. It does look like they are part of the public API as return type of `client::connect`?

src/client/mod.rs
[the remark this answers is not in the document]
sanmai-NL: From my perspective, this detail had an effect on my review (clarity remark), and it was also relevant when I came up with an alternative to the change you did make around here, as part of the review. That ended up in this corner though: https://github.com/djc/tokio-imap/pull/14#discussion_r167402599, hidden by default, accessible via https://github.com/djc/tokio-imap/pull/14/files/1f4d9740d9a9c604d95a5eec5a08c7197c0b5fc0#diff-77939fc9650773355cff92fe8aef3e90.

djc:
It looks like all the places where you commented on the old version (or maybe, where you commented on something not part of the diff) got kind of confused. I did find them in the end in the changes, so was then able to respond to them. Still weird that you have "approved" and "requested changes" so many times.

src/client/mod.rs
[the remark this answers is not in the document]
sanmai-NL: `poll()` is quite a hotspot I presume. This alternative for me is bit easier to follow because of lack of the extra binding and unwraps.

src/client/mod.rs
djc: `ResponseStream` consumes the transport while it is in the process of sending the request. This could be worked around by adding a buffer, which might improve the situation a bit but will add more errors to be thrown if the buffer is full. In this sense the current API makes it statically impossible to send while no buffer is available, which I like.

src/client/mod.rs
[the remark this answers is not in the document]
djc: I can still put a comment in there for contributors, that doesn't mean they have to be documented. As far as I'm concerned, the public type of `ImapConnectFuture` is `Future<Item = (TlsClient, ResponseData), Error = io::Error>`, so I don't want to document any part of the internals.

sanmai-NL:
Yeah I was in doubt about the UI since I haven’t done a complete review via GitHub before. I figured that comments should be made line-by-line within a commit, then the commit as a whole should be reviewed, in the form of a conclusion and optionally a general remark. It seems as if only a whole PR can be approved/etc. but that is strange since you can add commits and rewrite history all the time as PR owner.

src/client/mod.rs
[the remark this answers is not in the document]
djc: I want it in a separate PR if only so GitHub presents me with a proper diff so I can analyze the differences. 😄 And still, presumptions are often misguided when it comes to performance. Plus, this is the `ImapConnectFuture`, so it's much less likely to be a hotspot compared to anything related to the `ResponseStream` -- it's called maybe 5 times over the lifetime of a connection.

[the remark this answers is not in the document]
sanmai-NL: We can benchmark after merging this PR, indeed, if you feel like that.

djc:
Rebased as eb5bb04..60ac46c.

PR #17, opened 2018-05-06
Fix parsing of mailbox names such as `INBOX.Tests`

mcarton:
Didn't see the point of that extra `INBOX` special case, and the subdirectories of my main mail account were formatted as `INBOX.Foo`.

### DIFF
--- a/imap-proto/src/parser.rs
+++ b/imap-proto/src/parser.rs
@@ -127,7 +127,6 @@ named!(astring<&[u8]>, alt!(
 ));
 
 named!(mailbox<&str>, alt!(
-    map!(tag_s!("INBOX"), |s| "INBOX") |
     map_res!(astring, str::from_utf8)
 ));
 
@@ -766,6 +765,22 @@ mod tests {
         match parse_response(b"* 4 FETCH (UID 71372 RFC822.HEADER {10275}\r\n") {
             IResult::Incomplete(nom::Needed::Size(10319)) => {},
             rsp => panic!("unexpected response {:?}", rsp),
+        }
+    }
+
+    #[test]
+    fn test_list() {
+        match ::parser::mailbox(b"INBOX.Tests") {
+            IResult::Done(_, mb) => {
+                assert_eq!(mb, "INBOX.Tests");
+            },
+            rsp @ _ => panic!("unexpected response {:?}", rsp),
+        }
+
+        match parse_response(b"* LIST (\\HasNoChildren) \".\" INBOX.Tests\r\n") {
+            IResult::Done(_, Response::MailboxData(mb)) => {
+            },
+            rsp @ _ => panic!("unexpected response {:?}", rsp),
         }
     }
 }


djc:
Thanks for the patch! Unfortunately in my reading of the spec there is some special handling for `INBOX`, so rather than remove the special case I think we should probably firm that up here.

I've looked into it a bit, and this is what I came up with:

https://github.com/djc/tokio-imap/commit/badc7a78cb769b4b5b332b5dd8b052b1607a7ffc

What do you think? @jonhoo, do you mind taking a look as well? This will require a new minor release, since it introduces the use of `Cow<str>` in some of the data types.

jonhoo:
I think `.to_uppercase` is probably overkill here. Use [`eq_ignore_ascii_case`](https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case) instead -- that way the API can stay the same :)

djc:
Hmm, nice suggestion. In my understanding that is not fully compliant with the spec, though:

```
mailbox         = "INBOX" / astring
                    ; INBOX is case-insensitive.  All case variants of
                    ; INBOX (e.g., "iNbOx") MUST be interpreted as INBOX
                    ; not as an astring.  An astring which consists of
                    ; the case-insensitive sequence "I" "N" "B" "O" "X"
                    ; is considered to be INBOX and not an astring.
                    ;  Refer to section 5.1 for further
                    ; semantic details of mailbox names.
```

jonhoo:
Oh, I see, you specifically also want to *yield* `INBOX`. That seems fine, as you can simply use a literal `"INBOX"`, which is `&'static str` (and thus conforms with any lifetime). It might also be that you'd want to do something like `opt!(tag_no_case!("INBOX"))` for matching `INBOX`.

As a separate question, I can't quite figure out what the spec dictates for something like `iNbOx.foo`. Should that be returned as `INBOX.foo`, or left alone?

djc:
Nice, that's much better. I've merged this, with this additional change:

```diff
diff --git a/imap-proto/src/parser.rs b/imap-proto/src/parser.rs
index 50ba0d0..8d17453 100644
--- a/imap-proto/src/parser.rs
+++ b/imap-proto/src/parser.rs
@@ -126,8 +126,15 @@ named!(astring<&[u8]>, alt!(
     string
 ));
 
-named!(mailbox<&str>, alt!(
-    map_res!(astring, str::from_utf8)
+named!(mailbox<&str>, map!(
+    map_res!(astring, str::from_utf8),
+    |s| {
+        if s.eq_ignore_ascii_case("INBOX") {
+            "INBOX"
+        } else {
+            s
+        }
+    }
 ));
 
 named!(flag_extension<&str>, map_res!(
@@ -770,9 +777,9 @@ mod tests {
 
     #[test]
     fn test_list() {
-        match ::parser::mailbox(b"INBOX.Tests") {
+        match ::parser::mailbox(b"iNboX") {
             IResult::Done(_, mb) => {
-                assert_eq!(mb, "INBOX.Tests");
+                assert_eq!(mb, "INBOX");
             },
             rsp @ _ => panic!("unexpected response {:?}", rsp),
         }
```

As I read the spec, only the "INBOX" case is handled differently:

```
   In particular, this specification takes no position on case
   sensitivity in non-INBOX mailbox names.  Some server implementations
   are fully case-sensitive; others preserve case of a newly-created
   name but otherwise are case-insensitive; and yet others coerce names
   to a particular case.  Client implementations MUST interact with any
   of these.  If a server implementation interprets non-INBOX mailbox
   names as case-insensitive, it MUST treat names using the
   international naming convention specially as described in section
   5.1.3.
```

(That is, I'm assuming "non-INBOX mailbox names" include those like "INBOX.foo".)

PR #22, opened 2018-07-17
Upgrade nom to 4.0

theduke:
Note: Relevant breaking changes:

* IResult is now an alias for std Result,
  with Incomplete being a Err(nom::Err::Incomplete)

* Incomplete now contains only the additional needed size
  instead of the total expected size

There is **one failing test** that I couldn't figure out right away.

### DIFF
--- a/imap-proto/Cargo.toml
+++ b/imap-proto/Cargo.toml
@@ -15,4 +15,4 @@ maintenance = { status = "actively-developed" }
 travis-ci = { repository = "djc/tokio-imap" }
 
 [dependencies]
-nom = "3.1"
+nom = "4.0"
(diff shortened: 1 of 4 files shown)


theduke:
Figured out the failing test.
`astring` parser was failing because `take_while1` now produces `Incomplete` when reaching EOF.

djc:
Thanks for looking at this! I had all the trivial stuff finished but had hoped there would be a nom-4.0.1 that didn't need the extraneous `ErrorKind` import...

djc:
Oh, and since you appear to actually use tokio-imap, I'd love to hear how you're using it/what for!

PR #25, opened 2018-08-24
Add workaround for two consecutive spaces in response

vandenoever:
This is an issue with the DavMail server.

### DIFF
--- a/imap-proto/src/parser.rs
+++ b/imap-proto/src/parser.rs
@@ -528,6 +528,7 @@ named!(msg_att_rfc822<AttributeValue>, do_parse!(
 
 named!(msg_att_rfc822_header<AttributeValue>, do_parse!(
     tag_s!("RFC822.HEADER ") >>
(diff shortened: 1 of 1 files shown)


djc:
The change looks good! You committed this against 0.4.1, can you rebase on top of master please?

vandenoever:
Done

djc:
Your test fails, I guess you're using nom 3 stuff instead of nom 4? Can you fix that up?

vandenoever:
Right. I'll do so later today. I assumed it would stay the same.

djc:
Here's an example fix from the nom 3 -> 4 conversion:

https://github.com/djc/tokio-imap/commit/c7ed8c8f5c905b45799768104e503fc505b83c9e#diff-c7845db764eb3d1f65eee26827f7fcc5R773

vandenoever:
It seems to work now, locally and in the CI.

djc:
Thanks!

PR #28, opened 2018-10-21
imap-proto/parser: add HIGHESTMODSEQ status attribute

dario23:
the HIGHESTMODSEQ attribute is specified in
https://tools.ietf.org/html/rfc4551. although this doesn't aim to
implement that rfc to any level of completion, i needed this basic level
of support because some servers (dovecot in my case) always send
HIGHESTMODSEQ-responses with STATUS lines

### DIFF
--- a/imap-proto/src/parser.rs
+++ b/imap-proto/src/parser.rs
@@ -364,6 +364,7 @@ named!(mailbox_data_lsub<Response>, do_parse!(
 // so that it can return a valid enum object instead of just a key.
 named!(status_att<StatusAttribute>, do_parse!(
     key: alt!(
+        tag_s!("HIGHESTMODSEQ") |
         tag_s!("MESSAGES") |
         tag_s!("RECENT") |
         tag_s!("UIDNEXT") |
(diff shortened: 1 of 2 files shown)


djc:
Makes sense, thanks for patching it in! Rebased as 5bd95ca205de1c3c295f165f3813c46ee3013f91 and applied a tiny bit of polish in 715eaeff6333a07be0d3a5084db23f6f2b4bf6f5. (I've also taken care of the problems that made CI fail.)

dario23:
just noticed that in my PR `AttributeValue::HighestModSeq` has an `u32` value, whereas [rfc 4551](https://tools.ietf.org/html/rfc4551#section-4) and the other code say it should be `u64`. i'm not sure how to reopen a PR/if i'm allowed to do that at all, but maybe you could just commit the fix yourself or cherry-pick from [my repo](https://github.com/dario23/tokio-imap/commit/2d760f1925b5e685c2fb2c2516dce3ee97dc6180) :-)

djc:
Easiest for me would be if you can rebase your fix on top of master and submit a new PR would you mind doing that?

djc:
Note that you probably also need to change the parser for this.

dario23:
#29 :-)

PR #29, opened 2018-10-21
imap-proto/parser: correct HighestModSeq type to u64

dario23:
fix for highestmodseq being u64, see https://github.com/djc/tokio-imap/pull/28#issuecomment-431669527


djc:
Great, thanks!

Issue #27, opened 2018-09-30
`quoted_data` should return `Needed::Unknown` until no unescaped " found


duesee:
I probably noticed a bug in the `quoted` / `quoted_data` parser. Because I am just  becoming acquainted with your code and with nom, I am not sure if I misunderstand something. Here is my failing test:

```rust
#[test]
fn test_quoted() {
    // This is fine.
    assert!(quoted(b"\"I am finished!\"").is_ok());

    // Here, the parser can't know if more data will follow...
    assert_eq!(
        quoted(b"\"I am fini..."), // ...because there is no closing quote.
        Err(Err::Incomplete(Needed::Unknown))
    );
}
```

djc:
Oh, yeah, you're probably right. It's likely that all the parsers written as normal Rust functions instead of using nom combinators have problems like this. So yes, I agree with your assessment. Are you in a position to fix it yourself?

duesee:
Currently, I am trying to find out why some fetch responses are not parsed correctly.
I could commit the test first so that we don't forget about it and fix it later when I'm done with the fetch issue.

By the way: I also did some minor refactorings to group related parsers more closely together so that I can find them better. I also find it useful to distinguish between "core primitives" like `atom`, `astring`, etc. and the "real imap types". If your interested in those changes as weill, I will open a pull request.

djc:
Sure, you can open a PR and we can discuss the best way forward!

djc:
@duesee how did you fare with this? If you still have some unfinished stuff somewhere, I'd like to be able to look at it!

duesee:
Hey, I started implementing some missing parsers, then implemented an ABNF parser in Rust/Nom in the hope to be able to automate the manual process of typing the Nom parsers. To cut the long story short, I got distracted by many other stuff and didn't focused on it for a while :see_no_evil:  I will take a look into it the next days and see if I can at least fix this issue.

djc:
Hah, sounds like a proper yak shaving session! No pressure if you no longer want to work on this, was just wondering about the status.

duesee:
I opened a pull request to merge some of the refactoring and the tests. BTW, what do you think about the idea to generate the parsers? Do you think this can be accomplished?

PR #37, opened 2019-01-12
Progress with #27

duesee:


djc:
So, I think this is an okay direction to take things in, but if you put all that in a single commit I will not be able to carefully review your changes. If you want to get this merged, it needs to be small commits that are easy to review for correctness (like, one for adding comments, one for splitting into modules, one for reformatting).

I also don't think having large swaths of commented out grammar is useful. imap-proto doesn't need to implement all of the RFC's grammar to be useful.

sanmai-NL:
An observation. The commented out grammar does signal which part of the RFC was implemented and which wasn't yet. The general idea behind that I can understand, since I also found it somewhat challenging to map the RFC to this crate's code.

djc:
Yes, I understood what the commented out code was trying to do. I don't agree that it is useful. Adding the RFC grammar should already help mapping the RFC to the code.

duesee:
Yeah, not a very nice PR... I will split it :-)

The comments doesn't look very nice, yes. But they will be useful for someone implementing the remaining parsers. Especially, as the RFC is lexicographically sorted and not by relation.

There are several "clusters" of ABNF defs., which could even be used to determine what goes in which file. So maybe adding all of them is even better? This would make it very clear what (and what not) is implemented and how.

PR #39, opened 2019-02-11
 Split parsers and fix parsing of rfc3501 test envelope 

marcelbuesing:
Hi this may be a little controversial but I split the parsers into (rust) sub modules. I personally think it's difficult to keep track where they are actually defined otherwise. Furthermore I added some comments and tests.
I ran into an issue with the `msg_att_envelope` parser using the example provided in rfc3501 and fixed that. You can reproduce the issue when running the added tests and undoing the change in `msg_att_envelope`. Since I currently have some free time I'm planning to add more tests. 

Btw. any idea what would be a good way to track what parsers of the RFCs have not been implemented yet, maybe a simple markdown checklist of the Formal Syntax chapters? 

I wish it would be possible to link the parsers code documentation to the RFC Formal Syntax definition e.g. 
```
\\\ [rfc3501 - Formal Syntax - address](https://tools.ietf.org/html/rfc3501#section-9#address)
named!(address<Address>, do_parse!(
```

### DIFF
--- a/imap-proto/src/parser/mod.rs
+++ b/imap-proto/src/parser/mod.rs
@@ -1,0 +1,4 @@
+pub use self::rfc3501::*;
+
+mod rfc3501;
+mod rfc4551;
--- a/imap-proto/src/parser/rfc3501.rs
+++ b/imap-proto/src/parser/rfc3501.rs
@@ -322,7 +322,7 @@ named!(opt_addresses<Option<Vec<Address>>>, alt!(
     map!(tag_s!("NIL"), |_s| None) |
     do_parse!(
         tag_s!("(") >>
-        addrs: many1!(address) >>
+        addrs: separated_nonempty_list!(opt!(tag!(" ")), address) >>
         tag_s!(")") >>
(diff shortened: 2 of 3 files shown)


djc:
Seems broadly okay to me. Can you make the minor changes suggested? Also bonus points for having a trailing newline at the end of each file.

As a tracking device, maybe add a markdown list in the module comment of each item named in the formal syntax for that spec/module?

marcelbuesing:
Might not have been a particular clever idea to force push. Anyway I renamed: `rfc4551::status_att_val` to `rfc451::status_att_val_highest_mod_seq`. Removed the redundant space. Added newlines to the end of the files. Updated the comments in rfc4551. Removed this part in `mod.rs`:

```
#![cfg_attr(rustfmt, rustfmt_skip)]
#![cfg_attr(feature = "cargo-clippy", allow(redundant_closure))]
```

marcelbuesing:
Regarding the tracking, that sounds like a good suggestion.

djc:
Thanks!

PR #41, opened 2019-04-16
Implemented missing response codes.

Emm54321:
Implemented ALERT, BADCHARSET, CAPABILITY, and PARSE response codes. See section 7.1 of rfc 3501.

### DIFF
--- a/imap-proto/src/parser/rfc3501.rs
+++ b/imap-proto/src/parser/rfc3501.rs
@@ -116,9 +116,20 @@ named!(resp_text_code_badcharset<ResponseCode>, do_parse!(
     (ResponseCode::BadCharset(ch))
 ));
 
+named!(capability_list<Vec<&str>>, do_parse!(
+    capabilities1: many_till!(capability, tag_s!(" IMAP4rev1")) >>
+    capabilities2: many0!(capability) >> ({
+        let mut v = Vec::with_capacity(10);
+        v.extend(capabilities1.0);
+        v.push("IMAP4rev1");
+        v.extend(capabilities2);
+        v
+    })
+));
+
 named!(resp_text_code_capability<ResponseCode>, do_parse!(
     tag_s!("CAPABILITY") >>
-    capabilities: many1!(capability) >>
+    capabilities: capability_list >>
     (ResponseCode::Capabilities(capabilities))
 ));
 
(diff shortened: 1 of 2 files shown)


imap-proto/src/parser/rfc3501.rs
```diff
@@ -97,6 +97,12 @@ named!(resp_text_code_alert<ResponseCode>, do_parse!(
     (ResponseCode::Alert)
 ));
 
+named!(resp_text_code_capability<ResponseCode>, do_parse!(
+    tag_s!("CAPABILITY") >>
+    capabilities: many1!(capability) >>
```
djc: It looks like technically this requires that the list of capabilities includes `IMAP4rev1`, maybe we should error out if it does not?

imap-proto/src/parser/rfc3501.rs
```diff
@@ -727,5 +747,27 @@ mod tests {
...
+        match parse_response(b"* NO [BADCHARSET] error\r\n") {
+            Ok((_, Response::Data {
+                status: Status::No,
+                code: Some(ResponseCode::BadCharset(None)),
+                information: Some("error")
+            })) => {},
+            rsp @ _ => panic!("unexpected response {:?}", rsp)
+        }
+
+        match parse_response(b"* NO [BADCHARSET (utf-8 latin1)] error\r\n") {
```
djc, line 795: Would you mind adding a test here that has an empty list of charsets?

imap-proto/src/parser/rfc3501.rs
Emm54321: I'm not sure such a test belongs to the parser. The client may want to try using a server that doesn't include IMAP4rev1 in its capabilities.

imap-proto/src/parser/rfc3501.rs
Emm54321: Ok, I added it.

imap-proto/src/parser/rfc3501.rs
djc: The formal syntax in the RFC requires that it's there, so I think that makes it a parser-level check, wouldn't you agree?

Emm54321: Indeed, I hadn't noticed it was part of the grammar.

djc:
Thanks! I'll probably simplify the capabilities code a bit later on.

PR #43, opened 2019-04-19
NIL only parsed as literal in nstring

avitex:
As far as I can see from [RFC3501](https://tools.ietf.org/html/rfc3501#page-88) `nil`'s only possible value is a literal `NIL`, however is currently parsed as both a literal, and a quoted string in `nstring`.

This PR allows nil to be parsed only as a literal.


djc:
Huh, I wonder how that got in there. Thanks for fixing it!

Issue #42, opened 2019-04-17
Parse `FETCH ... BODYSTRUCTURE` response


lberezy:
https://github.com/djc/imap-proto/issues/15
https://github.com/djc/imap-proto/pull/16

I was wondering if there was any plans to merge/support BODYSTRUCTURE responses as was discussed when this crate was [imap-proto](https://github.com/djc/imap-proto). It would be useful for upstream crates (such as [rust-imap](https://github.com/jonhoo/rust-imap)).

djc:
I'll see if I can polish up that PR and merge it soon.

avitex:
@djc I'm currently working on a PR to fully parse `BODYSTRUCTURE` in my fork.

djc:
Even better, thanks!

PR #44, opened 2019-04-20
Implement `BODYSTRUCTURE`

avitex:
Still working on this for #42, not yet ready to merge.
Just open to comments at this point.

On a side note, @djc, would you like me to split out changes to core.rs in a separate PR?
Just a few changes for refactoring available if you like.

TODO:
- [x] Impl message
- [x] Impl multipart
- [x] More tests

### DIFF
--- a/imap-proto/src/core.rs
+++ b/imap-proto/src/core.rs
@@ -59,6 +59,12 @@ pub fn quoted_data(i: &[u8]) -> IResult<&[u8], &[u8]> {
     Ok((&i[len..], &i[..len]))
 }
 
+/// space = a single space character
+named!(pub space, tag_s!(" "));
+
+// nil = "NIL"
+named!(pub nil, tag_s!("NIL"));
+
 /// quoted = DQUOTE *QUOTED-CHAR DQUOTE
 named!(pub quoted<&[u8]>, do_parse!(
     tag_s!("\"") >>
@@ -81,29 +87,28 @@ named!(pub literal<&[u8]>, do_parse!(
 /// string = quoted / literal
 named!(pub string<&[u8]>, alt!(quoted | literal));
 
+/// string bytes as as utf8
+named!(pub string_utf8<&str>, map_res!(string, str::from_utf8));
+
 /// nstring = string / nil
 named!(pub nstring<Option<&[u8]>>, alt!(
-    map!(tag_s!("NIL"), |_| None) |
+    map!(nil, |_| None) |
     map!(string, |s| Some(s))
 ));
 
-named!(pub string_utf8<&str>, map!(string, |s| str::from_utf8(s).unwrap()));
-
-named!(pub nstring_utf8<Option<&str>>, map!(nstring, |val| val.map(|s| str::from_utf8(s).unwrap())));
+/// nstring bytes as utf8
+named!(pub nstring_utf8<Option<&str>>, alt!(
+    map!(nil, |_| None) |
+    map!(map_res!(string, str::from_utf8), |s| Some(s))
+));
 
 /// number          = 1*DIGIT
 ///                    ; Unsigned 32-bit integer
 ///                    ; (0 <= n < 4,294,967,296)
-named!(pub number<u32>, map_res!(
-    map_res!(nom::digit, str::from_utf8),
-    str::parse
-));
+named!(pub number<u32>, flat_map!(nom::digit, parse_to!(u32)));
 
 /// same as `number` but 64-bit
-named!(pub number_64<u64>, map_res!(
-    map_res!(nom::digit, str::from_utf8),
-    str::parse
-));
+named!(pub number_64<u64>, flat_map!(nom::digit, parse_to!(u64)));
 
 /// atom = 1*ATOM-CHAR
 named!(pub atom<&str>, map_res!(take_while1_s!(atom_char),
--- a/imap-proto/src/core.rs
+++ b/imap-proto/src/core.rs
@@ -87,6 +87,10 @@ named!(pub nstring<Option<&[u8]>>, alt!(
     map!(string, |s| Some(s))
 ));
 
+named!(pub string_utf8<&str>, map!(string, |s| str::from_utf8(s).unwrap()));
+
+named!(pub nstring_utf8<Option<&str>>, map!(nstring, |val| val.map(|s| str::from_utf8(s).unwrap())));
+
 /// number          = 1*DIGIT
 ///                    ; Unsigned 32-bit integer
 ///                    ; (0 <= n < 4,294,967,296)
--- a/imap-proto/src/core.rs
+++ b/imap-proto/src/core.rs
@@ -62,7 +62,7 @@ pub fn quoted_data(i: &[u8]) -> IResult<&[u8], &[u8]> {
 /// space = a single space character
 named!(pub space, tag_s!(" "));
 
-// nil = "NIL"
+/// nil = "NIL"
 named!(pub nil, tag_s!("NIL"));
 
 /// quoted = DQUOTE *QUOTED-CHAR DQUOTE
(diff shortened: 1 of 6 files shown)


avitex:
@djc A couple additions to the above question:

- What's the reasoning behind putting `body.rs` outside the parser module?
- What do you think about the meda-types specified in the RFC to be in an enum, with a custom variant?
- What's the reasoning behind utf8 and non-utf8 strings? Does it make sense to make all utf8, or none at all?

I'm sure I'll think of some more soon. Also note body structure depends on envelope, which I've split up from the message attr.

Cheers!

avitex (commented):
I'm not sure I like what cargo-fmt has done here, looks very messy

imap-proto/src/types.rs
```diff
@@ -111,13 +111,101 @@ pub enum SectionPath {
     Part(Vec<u32>, Option<MessageSection>),
 }
 
+#[derive(Debug, Eq, PartialEq)]
+pub struct BodyParam<'a> {
```
avitex, line 115: @djc Would you prefer body param as is here, or as a tuple?

djc: I think a tuple is probably fine for this.

imap-proto/src/types.rs
```diff
@@ -111,13 +111,101 @@ pub enum SectionPath {
...
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum BodyExtension<'a> {
+    Num(u32),
+    Str(Option<&'a str>),
+    List(Vec<BodyExtension<'a>>),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct BodyStructureText<'a> {
```
djc, line 134: There is a lot of duplication between the different `BodyStructure` types. Can you think about a way to reduce the duplication here? Not yet sure what that would look like, but I wouldn't want to merge it as is.

imap-proto/src/types.rs
```diff
@@ -111,13 +111,101 @@ pub enum SectionPath {
...
+}
+
 #[derive(Debug, Eq, PartialEq)]
 pub enum AttributeValue<'a> {
     BodySection {
         section: Option<SectionPath>,
         index: Option<u32>,
         data: Option<&'a [u8]>,
     },
+    BodyStructure(Box<BodyStructure<'a>>),
```
djc, line 208: Why do we have the `Box` here?

imap-proto/src/core.rs
```diff
@@ -87,6 +87,10 @@ named!(pub nstring<Option<&[u8]>>, alt!(
     map!(string, |s| Some(s))
 ));
 
+named!(pub string_utf8<&str>, map!(string, |s| str::from_utf8(s).unwrap()));
```
djc: It'd be nice if you introduce these helpers to convert existing `str::from_utf8()` users that would benefit from these. Ideally in that case the whole change would go in a separate commit.

imap-proto/src/types.rs
```diff
@@ -111,13 +111,69 @@ pub enum SectionPath {
...
+        octets: u32,
+        md5: Option<&'a str>,
+        disposition: Option<BodyDisposition<'a>>,
+        lang: Option<Vec<&'a str>>,
+        loc: Option<&'a str>,
+        lines: u32,
+        extensions: Option<BodyExtension<'a>>
+    },
+    // TODO
+    Message,
```
djc: Please fold the commits that add these together into this one.

imap-proto/src/core.rs
```diff
@@ -59,6 +59,12 @@ pub fn quoted_data(i: &[u8]) -> IResult<&[u8], &[u8]> {
     Ok((&i[len..], &i[..len]))
 }
 
+/// space = a single space character
+named!(pub space, tag_s!(" "));
```
djc, line 63: What's the reason to extract this? It doesn't seem to be a separate syntax item in the RFC, and in this case it feels like the abstraction doesn't add much value beyond the 6-character reduction.

imap-proto/src/core.rs
```diff
@@ -59,6 +59,12 @@ pub fn quoted_data(i: &[u8]) -> IResult<&[u8], &[u8]> {
...
+
+// nil = "NIL"
+named!(pub nil, tag_s!("NIL"));
```
djc, line 66: I'd be hesitant about this, too, but since it's in the syntax it might be helpful.

imap-proto/src/core.rs
```diff
@@ -81,29 +87,28 @@ named!(pub literal<&[u8]>, do_parse!(
 /// string = quoted / literal
 named!(pub string<&[u8]>, alt!(quoted | literal));
 
+/// string bytes as as utf8
+named!(pub string_utf8<&str>, map_res!(string, str::from_utf8));
+
 /// nstring = string / nil
 named!(pub nstring<Option<&[u8]>>, alt!(
-    map!(tag_s!("NIL"), |_| None) |
+    map!(nil, |_| None) |
```
djc, line 95: But if you're going to convert existing uses, please make it into a separate commit.

imap-proto/src/core.rs
```diff
@@ -81,29 +87,28 @@ named!(pub literal<&[u8]>, do_parse!(
...
     map!(string, |s| Some(s))
 ));
 
-named!(pub string_utf8<&str>, map!(string, |s| str::from_utf8(s).unwrap()));
-
-named!(pub nstring_utf8<Option<&str>>, map!(nstring, |val| val.map(|s| str::from_utf8(s).unwrap())));
+/// nstring bytes as utf8
```
djc, line 99: Please don't mix code movement or reformatting in the same commit as functional changes, this makes it much harder for me to review.

djc (commented):
Thanks for all your work!

For a larger, more complex PR like this I'm going to be a bit nitpicky about how it's represented so I can do a proper review -- I hope that's okay. What I'd like to see here:

* Meaningful, coherent commits that make a small single change each, each of which passes the tests (including rustfmt)
* No flip-flopping: don't make small/non-functional changes in later commits to code from earlier commits

I recommend using interactive rebase to split and/or squash your existing commits, but given the sequence you got now it might be a bit tricky. In these cases sometimes I just start from scratch and redo the entire sequence (with cherry-picking of commits and some copy/paste work where it makes sense). Let me know if you need some tips on how to do that.

Otherwise it mostly looks good, although I really want to figure out a neater way of abstracting between the different body structure types.

avitex:
@djc Could you please have a look at my questions in my [above comment](https://github.com/djc/tokio-imap/pull/44#issuecomment-485221137).
I'm happy to pull out the changes from the core in a separate PR, made amends to this one and squash.

djc:
Enum for media types could make sense, what section of the RFC are you referring to exactly?

As for UTF-8 vs bytes, I think I was going for string types where it seems likely or is specified that no non-UTF-8 content will ever happen (I think some fields are specified as ASCII) and bytes for the rest (especially for large chunks that include email bodies).

avitex:
> Enum for media types could make sense, what section of the RFC are you referring to exactly?

There are a few different specific media-types specified in the [formal syntax](https://tools.ietf.org/html/rfc3501#section-9) - see `media-basic` for example

> As for UTF-8 vs bytes, I think I was going for string types where it seems likely or is specified that no non-UTF-8 content will ever happen (I think some fields are specified as ASCII) and bytes for the rest (especially for large chunks that include email bodies).

Seems fair enough; I'll have to have to look into more detail on the RFC for that.

djc:
IIRC there are crates around mime types already, maybe we should import their types rather than defining our own?

avitex:
Want to look at that as a separate PR?

djc:
Yeah, that would probably make sense.

avitex:
Closing in favour of #48. I have a dependency on this branch.

lberezy:
Closes #42

PR #45, opened 2019-04-26
Refactor core

avitex:
- Add `nil`.
- Add utf8 versions of `string`, `nstring`, `astring` and `quoted`.
- Simplify `number` and `number_64`.
- Used `delimited!` for `quoted`


djc:
Awesome, thanks!

PR #46, opened 2019-04-26
Refactor paren delimited structures with macros

avitex:


avitex:
@djc Off-topic, but I'm in your tokio-imap gitter now.

djc:
Ohh, this is a nice abstraction! Reading macro definitions is still a bit challenging for me, so maybe add some comments with examples for each expansion variant in a future PR?

avitex:
> Ohh, this is a nice abstraction! Reading macro definitions is still a bit challenging for me, so maybe add some comments with examples for each expansion variant in a future PR?

Will do!

PR #40, opened 2019-02-14
Update capabilities

marcelbuesing:
- Enforces `IMAP4rev1`in capability-data.
```
"capability-data = CAPABILITY" *(SP capability) SP "IMAP4rev1" *(SP capability)
```
- Turn capability into enum type.
- Added tests

- Clean up cargo doc comments from previous merge request in rfc4551. Still can't see them when generating docs via `cargo doc --open --document-private-items -p imap-proto` still it's consistent. I assume this is due to nom macro magic.

PR #47, opened 2019-04-26
Parse capabilities into enum

avitex:
Based off work by marcelbuesing (replaces #40).

### DIFF
--- a/imap-proto/src/parser/rfc3501.rs
+++ b/imap-proto/src/parser/rfc3501.rs
@@ -86,21 +86,9 @@ named!(resp_text_code_badcharset<ResponseCode>, do_parse!(
     (ResponseCode::BadCharset(ch))
 ));
 
-named!(capability_list<Vec<&str>>, do_parse!(
-    capabilities1: many_till!(capability, tag_s!(" IMAP4rev1")) >>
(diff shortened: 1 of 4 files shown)


imap-proto/src/parser/rfc3501.rs
```diff
@@ -169,16 +157,48 @@ named!(resp_text_code<ResponseCode>, do_parse!(
...
-    (atom)
+/// [RFC3501 - 9. Formal Syntax - capability](https://tools.ietf.org/html/rfc3501#section-9)
+named!(capability_auth<Capability>, do_parse!(
+    tag_s!("AUTH=") >>
+    auth: atom >>
+    (Capability::Auth(auth))
+));
+
+/// [RFC3501 - 9. Formal Syntax - capability](https://tools.ietf.org/html/rfc3501#section-9)
+named!(capability_atom<Capability>, do_parse!(
```
avitex: On second thought, both this and `capability_auth` could just be map!s in capability, there isn't really any need for them to be separate

avitex:
I've left the semver commit out for the bodystructure PR. Do you want a separate bump, or fine with this?

PR #48, opened 2019-04-26
Add BODYSTRUCTURE support

avitex:
Replaces #44

#### TODO
- Version bump

### DIFF
--- a/imap-proto/src/parser/rfc3501.rs
+++ b/imap-proto/src/parser/rfc3501.rs
@@ -319,7 +319,7 @@ named!(opt_addresses<Option<Vec<Address>>>, alt!(
     ), |v| Some(v))
 ));
 
-named!(envelope<Envelope>, paren_delimited!(
+named!(pub(crate) envelope<Envelope>, paren_delimited!(
     do_parse!(
         date: nstring_utf8 >>
         tag_s!(" ") >>
--- a/imap-proto/src/parser/rfc3501.rs
+++ b/imap-proto/src/parser/rfc3501.rs
@@ -319,29 +319,28 @@ named!(opt_addresses<Option<Vec<Address>>>, alt!(
     ), |v| Some(v))
 ));
 
-named!(msg_att_envelope<AttributeValue>, do_parse!(
-    tag_s!("ENVELOPE (") >>
-    date: nstring_utf8 >>
-    tag_s!(" ") >>
-    subject: nstring_utf8 >>
-    tag_s!(" ") >>
-    from: opt_addresses >>
-    tag_s!(" ") >>
-    sender: opt_addresses >>
-    tag_s!(" ") >>
-    reply_to: opt_addresses >>
-    tag_s!(" ") >>
-    to: opt_addresses >>
-    tag_s!(" ") >>
-    cc: opt_addresses >>
-    tag_s!(" ") >>
-    bcc: opt_addresses >>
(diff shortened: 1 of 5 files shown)


imap-proto/src/parser/rfc3501.rs
```diff
@@ -319,7 +319,7 @@ named!(opt_addresses<Option<Vec<Address>>>, alt!(
     ), |v| Some(v))
 ));
 
-named!(envelope<Envelope>, paren_delimited!(
+named!(pub (crate) envelope<Envelope>, paren_delimited!(
```
djc: Can we do `pub(crate)` here?

imap-proto/src/types.rs
```diff
@@ -118,13 +118,97 @@ pub enum SectionPath {
     Part(Vec<u32>, Option<MessageSection>),
 }
 
+pub type BodyParam<'a> = Option<Vec<(&'a str, &'a str)>>;
```
djc: I'd like the order here to be reversed: `BodyParam` at the bottom (no dependencies, used by the other types), and `BodyStructure` should come below `AttributeValue`. (See also https://deterministic.space/how-to-order-rust-code.html.)

imap-proto/src/types.rs
```diff
@@ -118,13 +118,97 @@ pub enum SectionPath {
...
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum BodyExtension<'a> {
+    Num(u32),
+    Str(Option<&'a str>),
+    List(Vec<BodyExtension<'a>>),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct BodyStructureText<'a> {
```
djc: As I mention in the previous PR, there's a lot of duplication here. Should probably factor out those fields that are on every variant, and probably also the fields that are on text/basic/message but not multipart. I'm not sure what good type names are to convey the concepts here, can you make a proposal?

|   | text | basic | message | multipart |
|---|---|---|---|---|
| param | x | x | x | x |
| disposition | x | x | x | x |
| lang | x | x | x | x |
| location | x | x | x | x |
| extension | x | x | x | x |
| id | x | x | x |   |
| description | x | x | x |   |
| encoding | x | x | x |   |
| octets | x | x | x |   |
| md5 | x | x | x |   |
| envelope | | | x | |
| lines | x | x |   |   |
| bodies | | | | x |
| media_type | | x |   |   |
| media_subtype | x | x |   | x |

djc (commented):
This was a lot easier to review than in the first round, thanks! Some things to work on below.

imap-proto/src/parser/rfc3501.rs
avitex: Removing the whitespace correct? There are couple of other files with the space between `pub` and `(crate)`

imap-proto/src/types.rs
avitex: How about the private types in [`body_structure.rs`](https://github.com/djc/tokio-imap/pull/48/commits/a5140492dc337b34cfc95f9c1af55609cd0ac21f#diff-2af6f4b98e4e6881c1c7dbdac302e407R9)?

djc: Externalizing BodyFields could make sense, but the duplication between 1Part and MPart doesn't make sense to me.

lberezy:
Closes #42

PR #49, opened 2019-05-03
Derive Hash for Capability enum

lberezy:
Required for use of Capability enumeration (recent PR https://github.com/djc/tokio-imap/pull/47)  in [rust-imap crate](https://github.com/jonhoo/rust-imap/blob/master/src/types/capabilities.rs#L34).


djc:
Thanks!

PR #50, opened 2019-05-08
Fix regressions

avitex:
nom does not seem to be able to determine when the address list finishes with the optional space with the below combination of parsers.

```
paren_delimited!(separated_nonempty_list!(opt!(char!(' ')), address))
```

This PR adds a workaround, with an associated test that currently fails on master.

While checking the specification, I noticed a space is not specified anywhere in envelope addresses fields.
Not sure if the current behavior is against spec to accommodate a specific server?
```
address         = "(" addr-name SP addr-adl SP addr-mailbox SP addr-host ")"
env-bcc         = "(" 1*address ")" / nil
env-cc          = "(" 1*address ")" / nil
env-from        = "(" 1*address ")" / nil
env-reply-to    = "(" 1*address ")" / nil
env-sender      = "(" 1*address ")" / nil
env-to          = "(" 1*address ")" / nil
```

### DIFF
--- a/imap-proto/src/macros.rs
+++ b/imap-proto/src/macros.rs
@@ -12,7 +12,7 @@ macro_rules! parenthesized_list (
         paren_delimited!($i, separated_list!(char!(' '), $submac!($($args)*)))
     });
     ($i:expr, $f:expr) => (
-        parenthesized_nonempty_list!($i, call!($f));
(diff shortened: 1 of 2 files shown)


djc:
I think this looks good, let's just merge it.

djc:
It looks good, let's just merge it. :)

PR #53, opened 2019-08-28
Upgrade to nom 5

marcusklaas:
This takes care of a number of deprecation messages.

While nom 5 uses functions at its core instead of macros, the macros still work and this PR leaves the use of macros in place.


djc:
Ohh, thanks for starting the work on this! Didn't even know the macros were still there.

marcusklaas:
Woops - I forgot to update the tokio-imap crate :sweat_smile:

I'll include the necessary changes in this PR.

djc:
Thanks!

PR #55, opened 2019-09-04
Make token matches case-insensitive

### LINKED ISSUE #54: Capability "IMAP4REV1" is not accepted as valid
Since PR #47 parser requires `IMAP4rev1` capability to be present. Some IMAP servers send `IMAP4REV1` string instead: https://www.communigate.com/pub/Guide_516.pdf, p. 149

It results in parse error, so I can't connect to such servers with [imap crate](https://github.com/jonhoo/rust-imap), which uses imap-proto parser.

According to [RFC3501](https://tools.ietf.org/html/rfc3501#section-9) "Except as noted otherwise, all alphabetic characters are case-insensitive.".

Thunderbird and mutt accept `IMAP4REV1`.

avitex:
Fixes #54

@link2xt would you be able to review this too?

### DIFF
--- a/imap-proto/src/body.rs
+++ b/imap-proto/src/body.rs
@@ -14,18 +14,14 @@ named!(pub section_part<Vec<u32>>, do_parse!(
     })
 ));
 
-named!(pub section_msgtext<MessageSection>, map!(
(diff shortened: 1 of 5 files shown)


djc:
Thanks!

PR #56, opened 2019-09-04
Add IMAP4REV1 test case

link2xt:
Test that "IMAP4rev1" capability is matched case-insensitive

See issue https://github.com/djc/tokio-imap/issues/54


djc:
Thanks!

PR #59, opened 2019-10-24
Parsing of the metadata response (RFC5464)

Sergey-Kitov:
Implementation of METADATA extension response parser

### DIFF
--- a/imap-proto/src/parser/rfc3501.rs
+++ b/imap-proto/src/parser/rfc3501.rs
@@ -13,6 +13,7 @@ use nom::IResult;
 use std::str;
 
 use crate::parser::rfc4551;
+use crate::parser::rfc5464::resp_metadata;
 use crate::types::*;
 use crate::core::*;
 use crate::body::*;
@@ -512,7 +513,8 @@ named!(response_data<Response>, do_parse!(
         mailbox_data |
         message_data_expunge |
         message_data_fetch |
-        resp_capability
+        resp_capability |
+        resp_metadata
     ) >>
     tag!("\r\n") >>
     (contents)
--- a/imap-proto/src/parser/rfc5464.rs
+++ b/imap-proto/src/parser/rfc5464.rs
@@ -1,0 +1,292 @@
+//!
+//! https://tools.ietf.org/html/rfc5464
+//!
+//! IMAP METADATA extension
+//!
+
+// rustfmt doesn't do a very good job on nom parser invocations.
+#![cfg_attr(rustfmt, rustfmt_skip)]
+#![cfg_attr(feature = "cargo-clippy", allow(redundant_closure))]
+
+use nom::IResult;
+
+use crate::core::*;
+use crate::types::*;
+use std::str;
+
+fn is_entry_component_char(c: u8) -> bool {
+    c < 0x80 && c > 0x19 && c != b'*' && c != b'%' && c != b'/'
+}
+
+enum EntryParseStage<'a> {
+    PrivateShared(usize),
+    Admin(usize),
+    VendorComment(usize),
+    Path(usize),
+    Done(usize),
+    Fail(nom::Err<&'a [u8]>),
+}
+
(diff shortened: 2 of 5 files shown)


djc:
Thanks for working on this!

Functionally this looks pretty okay. Would you mind reformatting it a bunch so that the style within the macro calls looks more like rustfmt-formatted code? Notably:

* Opening braces for `if` conditions go on the same line
* Max line length at 100 characters
* Closing brace for multi-line blocks on its own line (possibly with comma, but not at the last line of the block
* Spaces inside braces for structs on the same line

Sergey-Kitov:
Does it look better now?

imap-proto/src/parser/rfc3501.rs
```diff
@@ -13,6 +13,7 @@ use nom::IResult;
 use std::str;
 
 use crate::parser::rfc4551;
+use crate::parser::rfc5464::*;
```
djc: Okay, please don't use `*` imports -- sorry, should have mentioned that the first time around.

imap-proto/src/parser/rfc3501.rs
```diff
@@ -47,7 +48,7 @@ named!(status<Status>, alt!(
     status_bye
 ));
 
-named!(mailbox<&str>, map!(
+named!(pub(crate) mailbox<&str>, map!(
```
djc: Why make this `pub(crate)`? That seems unrelated to the main goal of this PR.

imap-proto/src/parser/rfc5464.rs
```diff
@@ -0,0 +1,273 @@
...
+    VendorComment(usize),
+    Path(usize),
+    Done(usize),
+    Fail(nom::Err<&'a [u8]>),
+}
+
+fn check_private_shared(i: &[u8]) -> EntryParseStage {
+    if i.starts_with(b"/private") {
+        return EntryParseStage::VendorComment(8);
+    }
```
djc: Using `else if` would be preferable here (saves one line and logically makes sense).

imap-proto/src/parser/rfc5464.rs
```diff
@@ -0,0 +1,273 @@
...
+    if i.starts_with(b"/shared") {
+        return EntryParseStage::Admin(7);
+    }
+    return EntryParseStage::Fail(
+        nom::Err::Error(
+        b"Entry Name doesn't start with /private or /shared"));
```
djc: Please put the closing parenthesis on the next line here.

imap-proto/src/parser/rfc5464.rs
```diff
@@ -0,0 +1,273 @@
...
+}
+
+fn check_admin(i: &[u8], l:usize) -> EntryParseStage {
+    if i[l..].starts_with(b"/admin") {
+        return EntryParseStage::Path(l+6);
```
djc: Please put spaces around the `+` here.

djc (commented):
There's a bunch of formatting issues still. I would probably enable formatting in the new module, format all the non-macro code and then manually format the macro code again (or find some other way to properly format the non-macro code).

imap-proto/src/parser/rfc3501.rs
Sergey-Kitov: There was some reason for making it pub(crate), but it looks like the reason has gone during development.

imap-proto/src/parser/rfc5464.rs
Sergey-Kitov: May be it is better to put error message on the same line as  "...Error("

djc: Please do whatever `rustfmt` does in this case.

Sergey-Kitov:
I have implemented all rustfmt corrections for non macro invocation code.

djc (approved):

djc:
Thanks!

Sergey-Kitov:
Thank you very much!

PR #68, opened 2019-12-19
fix(imap-proto): handle short continuation responses

### LINKED ISSUE #67: Handle short continuation response
Some servers (yandex for example) will only send `+\r\n` instead of `+ \r\n` as a continuation response, which the current parser does not accept.

dignifiedquire:
Fixes #67


djc:
Thanks!

PR #69, opened 2020-01-02
Add IDLE capability and parse it

seijikun:
Added IDLE to the parsed capabilities (see https://tools.ietf.org/html/rfc2177).

I don't know if `parser/rfc3501.rs` was the correct place to do it, or if it is wanted at all, since the capability is part of another rfc.

### DIFF
--- a/imap-proto/src/parser/rfc3501.rs
+++ b/imap-proto/src/parser/rfc3501.rs
@@ -161,6 +161,7 @@ named!(resp_text_code<ResponseCode>, do_parse!(
 
 named!(capability<Capability>, alt!(
     map!(tag_no_case!("IMAP4rev1"), |_| Capability::Imap4rev1) |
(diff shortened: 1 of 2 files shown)


djc:
Why does this even need to be parsed separately from the generic capability variants?

seijikun:
It doesn't need to. I just thought it was nicer to have a type-safe "constant" when checking if a server has a certain capability (https://docs.rs/async-imap/0.1.1/async_imap/types/struct.Capabilities.html), instead of checking by string.
Or is there something else I missed?

djc:
I think I decided on implementing it that having variants for each capability didn't make much sense, particularly since the full set isn't really enumerated anywhere, and the grammar isn't defined that way, either -- imap-proto largely follows the spec in terms of grammar (both RFC 3501 and just a few extensions). So I don't think it makes sense to implement this. Checking by string, in this case, seems fine to me, since that's going to be what's going on under the cover anyway.

PR #72, opened 2020-02-02
Parse HEADER.FIELDS

link2xt:
Currently [`HEADER.FIELDS` variant of `section-msgtext`](https://tools.ietf.org/html/rfc3501#page-89) is not handled at all.

### DIFF
--- a/imap-proto/src/body.rs
+++ b/imap-proto/src/body.rs
@@ -15,6 +15,12 @@ named!(pub section_part<Vec<u32>>, do_parse!(
 ));
 
 named!(pub section_msgtext<MessageSection>, alt!(
+    do_parse!(
+        tag_no_case!("HEADER.FIELDS") >>
+        opt!(tag_no_case!(".NOT")) >>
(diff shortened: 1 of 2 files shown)


imap-proto/src/body.rs
```diff
@@ -15,6 +15,11 @@ named!(pub section_part<Vec<u32>>, do_parse!(
 ));
 
 named!(pub section_msgtext<MessageSection>, alt!(
+    do_parse!(tag_no_case!("HEADER.FIELDS") >>
```
djc: I'd prefer to have the immediate elements of `do_parse!()` indented at one level, rather than this way.

(Of course this PR is also still failing the tests.)

link2xt:
Works for me now, some tests in https://github.com/deltachat/deltachat-core-rust/pull/1236 pass

djc:
Thanks!

djc:
I've released 0.10.2 which contains this fix.

PR #80, opened 2020-03-25
Refactoring modules into `parser::` and `types::` + minor cleanups

duesee:


duesee:
I also added an example, so that someone who is unsure wheather to use this library or not can see what it does.

imap-proto/src/types/mod.rs
```diff
@@ -8,6 +8,9 @@ pub enum AttrMacro {
     Full,
 }
 
+// TODO: why not derive(Copy)?
```
djc: No clue. If `derive` works, that would clearly be preferable! This might predate my/the author's understanding of trait derivation.

imap-proto/src/lib.rs
```diff
@@ -3,17 +3,13 @@ extern crate nom;
 
 #[macro_use]
 mod macros;
-mod parser;
 
-pub mod body;
-pub mod body_structure;
-pub mod builders;
-pub mod types;
+// Public API
```
djc, line 11: This commit has too many things going on. Please separate them into separate commits so I can review them separately. If your commit message has a list like this, it's a good sign that it should have each bullet as a separate commit (moving files and modules around can be all in one commit, I suppose).

imap-proto/src/parser/core.rs
```diff
@@ -1,5 +1,4 @@
-use nom::{self, IResult};
-use std::str;
+use nom::{character::streaming::digit1, IResult};
```
djc: It's not clear why `digit1` gets imported here, but `std::str` has to go (I do think it makes sense to import `digit1`, but doing both seems somewhat inconsistent). I'd prefer to keep the `std::str` import, unless you have a good argument why we shouldn't use it. (Potentially we should import `std::str::from_utf8()` directly if that's the only thing being used from `std::str`.)

imap-proto/src/lib.rs
djc: (Also should mention that I like the direction to avoid wildcard imports!)

imap-proto/src/lib.rs
```diff
@@ -7,9 +7,11 @@ mod macros;
 // Public API
 
 pub use crate::parser::{rfc3501::parse_response, rfc5464::resp_metadata, ParseResult};
+
+// TODO: move builders module later to command::CommandBuilder and response::ResponseBuilder.
```
djc, line 15: This is changing stuff that you introduced in the same PR. Please squash.

imap-proto/src/builders/command.rs
```diff
@@ -1,5 +1,7 @@
-use crate::quoted_string;
-use crate::types::{AttrMacro, Attribute, State};
+use crate::{
```
djc, line 1: You did a bunch of these earlier, this should be probably be in the same commit.

imap-proto/src/parser/rfc3501/body_structure.rs
```diff
@@ -4,7 +4,7 @@
 use crate::{
     parser::{
         rfc3501::envelope,
-        core::{nstring_utf8, number, string_utf8, nil}
+        core::*,
```
djc, line 7: I'd prefer to keep the explicit imports here.

imap-proto/src/parser/rfc3501/mod.rs
```diff
@@ -10,7 +10,7 @@
 use crate::{
     parser::{
         ParseResult,
-        core::{astring_char, astring_utf8, atom_char, atom, number, quoted_utf8, nil, nstring, nstring_utf8, text},
+        core::*,
```
djc, line 13: And here.

imap-proto/src/parser/rfc5464.rs
```diff
@@ -118,7 +117,7 @@ fn entry_name(i: &[u8]) -> IResult<&[u8], &[u8]> {
 }
 
 fn slice_to_str(i: &[u8]) -> &str {
-    str::from_utf8(i).unwrap()
+    std::str::from_utf8(i).unwrap()
```
djc, line 120: This also appears to belong in an earlier commit.

tokio-imap/src/client.rs
```diff
@@ -13,7 +13,7 @@ use tokio_util::codec::Decoder;
 
 use crate::proto::{ImapCodec, ImapTransport, ResponseData};
 use imap_proto::builders::command::Command;
-use imap_proto::{Request, RequestId, State};
+use imap_proto::types::{Request, RequestId, State};
```
djc, line 16: Should be fixed in the commit that introduced it.

imap-proto/src/parser/rfc3501/mod.rs
```diff
@@ -23,6 +23,46 @@ use crate::{
 pub mod body;
 pub mod body_structure;
 
+pub fn parse_response(msg: &[u8]) -> ParseResult {
```
djc: I don't want this hoisted up top on its own, we should reorder the entire module at once to flow from high-level to low-level.

imap-proto/src/parser/core.rs
```diff
@@ -49,7 +49,7 @@ pub fn quoted_data(i: &[u8]) -> IResult<&[u8], &[u8]> {
 }
 
 // quoted-specials = DQUOTE / "\"
-pub fn quoted_specials(c: u8) -> bool {
+pub fn is_quoted_specials(c: u8) -> bool {
```
djc, line 57: Good idea. 👍

imap-proto/src/parser/core.rs
```diff
@@ -53,16 +58,27 @@ pub fn is_quoted_specials(c: u8) -> bool {
...
-//            ; Number represents the number of CHAR8s
-named!(pub literal<&[u8]>, do_parse!(
-    tag!("{") >>
-    len: number >>
-    tag!("}") >>
-    tag!("\r\n") >>
-    data: take!(len) >> // FIXME: 0x00 is not allowed
-    (data)
-));
+/// literal = "{" number "}" CRLF *CHAR8
```
djc, line 61: I'd like to split this into two commits: one commit that converts `literal` to function-style, and one commit that makes it more strict. Otherwise, it makes reviewing stuff much harder.

djc (commented):
So, a bunch of feedback. Note that in general I like this work, and pretty much all of the feedback about how you're presenting it for review. On that front, I'm intentionally on the strict side here so that in the future reviewing will be quicker and we don't have to go round for round. I hope that's okay. Also if you want to push back on any feedback I give that's fine, let's discuss if you disagree!

duesee:
Thank you for the feedback! Makes all sense :-) There is only one problem... I have no idea how to split commits in a PR. Should I close this one first? I will try tomorrow evening.

imap-proto/src/lib.rs
```diff
@@ -1,36 +1,32 @@
...
-// Will return Err if the argument contains illegal characters.
-//
-// Relevant definitions from RFC 3501 formal syntax:
-//
-// string = quoted / literal [literal elided here]
-// quoted = DQUOTE *QUOTED-CHAR DQUOTE
-// QUOTED-CHAR = <any TEXT-CHAR except quoted-specials> / "\" quoted-specials
-// quoted-specials = DQUOTE / "\"
-// TEXT-CHAR = <any CHAR except CR and LF>
+/// Returns an escaped string if necessary for use as a "quoted" string per
```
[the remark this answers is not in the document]
duesee: Two points: 1) it is still semantically a docstring (even though it is private and might never be exported in the future) 2) CLion -- and this is the real reason I changed it -- is now able to collapse it ;-)

imap-proto/src/parser/rfc3501/mod.rs
duesee: Generally, I also prefer explicit imports. But core feels like a prelude as most modules will probably need nearly * from core anyway. I will revert it for now.

djc:
I usually use `git rebase -i` to do an interactive rebase. If you set that to `edit` the offending commit, you can then `git reset HEAD^` to reset the branch pointer to the previous commit while keeping your changes, so you can make new commits with `git commit -p` or similar.

duesee:
Okay, I think I learned how to squash, rearrange and split commits :-) See #81. I added 3 more commits here which you can review. Otherwise the new branch is identical to this one. Just so you don't have to review anything twice.

PR #81, opened 2020-03-27
API changes and cleanups (obsoletes #80)

duesee:


imap-proto/src/parser/core.rs
```diff
@@ -1,6 +1,4 @@
-use nom::{self, IResult};
-
-use std::str;
```
djc, line 3: I don't think we've come to agreement on the removal of the `std::str` import?

imap-proto/src/lib.rs
```diff
@@ -7,18 +7,16 @@ extern crate assert_matches;
 
 #[macro_use]
 mod macros;
-mod parser;
 
-pub mod body;
-pub mod body_structure;
+// Public API
```
djc: Sorry, but this still has too much going on. Don't do reformatting in the same commit as changing the public API, please, because it's hard for me to review.

imap-proto/src/parser/rfc3501/mod.rs
```diff
@@ -60,15 +64,15 @@ named!(mailbox<&str>, map!(
 
 named!(flag_extension<&str>, map_res!(
     recognize!(pair!(tag!("\\"), take_while!(atom_char))),
-    str::from_utf8
+    std::str::from_utf8
```
djc: Also these changes to add `std` before `str` belong in the other commit if we want them at all.

imap-proto/src/parser/rfc4551.rs
```diff
@@ -8,8 +8,10 @@
 // rustfmt doesn't do a very good job on nom parser invocations.
 #![cfg_attr(rustfmt, rustfmt_skip)]
 
-use crate::core::number_64;
-use crate::types::*;
+use crate::{
+    parser::core::*,
```
djc: I think we agreed to hold off on `core::*` imports.

imap-proto/src/parser/core.rs
```diff
@@ -1,4 +1,9 @@
-use nom::{character::streaming::digit1, IResult};
+use nom::{
+    bytes::streaming::{tag, take},
```
djc, line 3: This change doesn't make sense anymore because there are now a bunch of things being imported that don't get used.

imap-proto/src/parser/rfc3501/mod.rs
```diff
@@ -534,7 +534,6 @@ pub fn parse_response(msg: &[u8]) -> ParseResult {
     response(msg)
 }
 
-
```
djc: This doesn't belong in this commit. Maybe some earlier commit made changes here? Otherwise a separate commit for this is also fine.

djc (commented):
Bunch more nits about the presentation here. Getting closer though! Thanks for sticking with it.

djc:
(Also feel free to use `push -f` to overwrite the previous commits on a PR branch, rather than opening a new one.)

duesee:
Okay, I further split the changing api commit. I am not sure why, but when I reverted the std::str change, I always ended up in conflicting merge hell.

imap-proto/src/parser/core.rs
duesee: Yes. I forgot it. See last comment.

djc:
So I have taken 7 of your commits, applied minor fixes and pushed them to master. Please try to rebase this branch on master and clean up the commit history. I think I want the following commits:

* Reformat imports everywhere (reformatting only, please)
* Import the nom `digit1` function directly
* Move modules around in the hierarchy/filesystem
* Move `ParseResult`
* Add `is_` prefix to predicates (remove the `std::str` addition from that commit)
* Reorder parsers and add visual sections (remove the change to the other file)
* Expand the `core::*` imports (don't forget to reformat after this)
* Add `is_char()` and use it to fix `is_atom_char()` and `is_text_char()`
* Replace `literal` with strict version

This really is highly similar to what you've got now, but there's still quite a bit of noise in there. Fixing up the `std::str` stuff in a later commit is too ugly for me in this case, sorry. :( You should be able to squash that commit into one of the earlier ones to fix up most of the issues.

duesee:
Sorry, just to be clear, is there anything you do not want? Not by means of commit presentation, but the end result?

djc:
I don't think there's anything I don't want here, but it'll be easier to be firm on that once I can review it more closely because it is in a more easily reviewable state.

djc:
Current patches are looking great, would be happy to merge once you clean up the formatting!

(You might like this one-liner: `git filter-branch --tree-filter 'cargo fmt' -- origin/master...`.)

djc:
In fact, I figured I might as well do that for you -- this has been merged. Thanks!!

PR #82, opened 2020-03-28
Improve ordering of parsers in core

duesee:
Hey, this was still missing. I hope it is not too exhausting to verify this kind of changes, but I think this really improves the readability and will also help me in a later stage, when I try to merge more stuff.


djc:
Thanks! Ideally in the future we make it so that groups that depend on some other group come before that other group (so `astring` and `nstring` should probably come before `string`).

PR #85, opened 2020-03-29
Add RFC 5161 ENABLED response parser and test.

mordak:
I needed this to handle the response for 'ENABLE QRESYNC'.


djc:
Thanks! I cleaned up the redundant closures in 296262a74eea6a228c37ef45fe043b4917f290d8.

PR #87, opened 2020-04-05
Improve quoted parser

duesee:


duesee:
"Please provide an upload token from codecov.io with valid arguments"

djc:
You can ignore the missing token, it doesn't get exposed to PRs from external contributors for security. Can we use a closure to pass to `take_while1()`? That would prevent the unwieldily-named inline function.

djc:
Nice, thanks!

PR #91, opened 2020-08-27
Add workaround for Zoho Mail invalid FLAGS response

link2xt:
imap.zoho.com server sends `\*` in FLAGS response, which is only allowed
in PERMANENTFLAGS list.

As a workaround, allow `\*` in FLAGS response and let the user interpret
the response.


djc:
Looks good, thanks. If you can make some noise towards Zoho about this issue, that would be nice!

PR #101, opened 2020-11-09
Add support for RFC7162 (QRESYNC) VANISHED response.

mordak:
RFC7162 defines a `VANISHED` response to the `EXPUNGE` command which must be used whenever a client has enabled `QRESYNC`. `VANISHED` may also appear in response to `SELECT/EXAMINE` or `UID FETCH`.

https://tools.ietf.org/html/rfc7162#section-3.2.10

The `VANISHED` response sends a sequence set of `UID`s indicating `UID`s which have been deleted from the mailbox.

I had to add parsers for `sequence-range` (eg `15:30`) and `sequence-set` (eg. `1,10:20,25,28:31`), and I put them in core.rs since they are actually defined in RFC 3501 and show up in a couple of other places, so may be generically useful.

I am not sure if the choice to return a `Vec<RangeInclusive>` is to your liking. Alternatives could be just a tuple of `(start, end)`, or expand the range out into a big list of `Vec<u32>`. I experimented with both options and settled on the `Vec<RangeInclusive>` because:

- It is literally how the spec defines them
- It is impossible to misinterpret, whereas a user would need to read the RFC to know `(start, end)` is inclusive
- It is efficient in that large ranges can be represented and iterated with low memory overhead.

The downside is it can be a bit awkward to use, since you end up iterating a list of iterators. Of course, the nicest way to represent this for the user is to expand the sequence set out into a `Vec<u32>`, but this could be wasteful / dangerous in some circumstances.

### DIFF
--- a/imap-proto/src/parser/core.rs
+++ b/imap-proto/src/parser/core.rs
@@ -38,6 +38,22 @@ pub fn number_64(i: &[u8]) -> IResult<&[u8], u64> {
     }
 }
 
+// seq-range       = seq-number ":" seq-number
+//                    ; two seq-number values and all values between
+//                    ; these two regardless of order.
+//                    ; seq-number is a nz-number
+pub fn sequence_range(i: &[u8]) -> IResult<&[u8], std::ops::RangeInclusive<u32>> {
+    map(tuple((number, tag(":"), number)), |(s, _, e)| s..=e)(i)
+}
+
+// sequence-set    = (seq-number / seq-range) *("," sequence-set)
+//                     ; set of seq-number values, regardless of order.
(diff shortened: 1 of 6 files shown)


djc:
This is great, thanks! Would it be helpful to get a release including this?

mordak:
> Would it be helpful to get a release including this?

A release including this would be helpful, yes! :-)

djc:
0.12.0 is out now.

mordak:
> 0.12.0 is out now.

Thank you! Just opened the domino PR for rust-imap: https://github.com/jonhoo/rust-imap/pull/172

And then my IMAP <-> Maildir sync engine will actually build against public crates!

Issue #34, opened 2018-11-22
Don't choke on empty SEARCH response


jonhoo:
```
C: a3 SEARCH ALL
S: * SEARCH
thread 'inbox' panicked at 'called `Result::unwrap()` on an `Err` value: Parse(Invalid([42, 32, 83, 69, 65, 82, 67, 72, 32, 13, 10]))'
```
I think this response should be valid, but imap-proto seems to choke on it?

djc:
That response is technically invalid per the RFC: the relevant `mailbox-data` variant is `"SEARCH" *(SP nz-number)`, which implies that a space can only follow `SEARCH` if it's followed by an ID. In your test, `SEARCH` has a trailing space but no following IDs. If you got that in the wild, we should probably make the parser more lenient.

(There's already a test case for a valid empty `SEARCH` response.)

jonhoo:
I got that using the [GreenMail mail test server](http://www.icegreen.com/greenmail/). Filed https://github.com/greenmail-mail-test/greenmail/issues/264.

PR #108, opened 2020-12-30
Allow trailing whitespace in search responses.

rbartlensky:
Hello! I have a small PR here to fix an issue I found in the wild. Let me know if I need to change anything about it!

This relates to #34.

While working on a project, I tried to connect to Yahoo's IMAP server,
and all SEARCH responses seemed to have a trailing whitespace. For
example in the case below, there is a whitespace (even though it might
not seem obvious) right after the last id:

```
C: a5 UID SEARCH ALL
S: * SEARCH 169654 169655 169656 169657 169658 169659 169660 169661
169662 169663 
```

I also happened to encounter this on empty SEARCH responses, which
led me to the issue linked above.


djc:
Thanks! I've added a comment about the spec deviation and published version 0.12.1 to crates.io.

rbartlensky:
@djc Thank you for the quick review!

PR #109, opened 2021-01-15
Parse UIDPLUS response codes

### LINKED ISSUE #66: Support parsing RFC 4315 (UIDPLUS Extension) response codes
[`UIDPLUS`] is useful and desired for implementing any kind of heavy-duty IMAP synchronization tool (e.g., a RIIR of offlineimap). This requires parsing of three additional response codes: `COPYUID`, `APPENDUID`, and `UIDNOTSTICKY`.

[`UIDPLUS`]: https://tools.ietf.org/html/rfc4315

mmirate:
I believe this would fix #66.

### DIFF
--- a/imap-proto/Cargo.toml
+++ b/imap-proto/Cargo.toml
@@ -15,6 +15,7 @@ edition = "2018"
 maintenance = { status = "passively-maintained" }
 
 [dependencies]
+either = "1.6.1"
 nom = { version = "6", default-features = false, features = ["std"] }
 
 [dev-dependencies]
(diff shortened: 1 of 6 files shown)


mmirate:
Seems your codecov.io integration is misconfigured?

djc:
Thanks for working on this!

I believe the codecov CI will not work for PRs from forks, due to not wanting to expose the token to external contributors.

This mostly looks good. I generally prefer not to rely on `Either` enums, instead defining our own with domain-specific variants. Additionally, please add some docstrings to the new parsers that link to the relevant RFC and show the syntax definition for that particular parser (as is done in many of the existing parsers).

mmirate:
... hmm. How do I monospace the ABNF without making cargo think it's a doctest?

djc (approved):

djc:
I'm fine with the syntax definitions without monospace. 😊

PR #110, opened 2021-01-15
Avoid running the codecov.io CI without the token

mmirate:
Unfortunately I have no idea how to test whether this actually works, other than sending in this PR and seeing what the CI does. And even that much, is assuming that CI will take into account the changes from this PR rather than just the repo.

~~EDIT: Oh. Huh. It doesn't run CI on PRs that don't contain any code changes. Makes sense, I suppose.~~

EDIT2: Wrong, see comment below.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,7 +59,7 @@ jobs:
 
   coverage:
     runs-on: ubuntu-latest
-    if: ${{ secrets && secrets.CODECOV_TOKEN }}
+    if: secrets && secrets.CODECOV_TOKEN
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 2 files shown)


djc:
Can you make some trivial change to a docstring or something? We can always revert that part later.

mmirate:
Done. Except, it's still not running CI? That's weird.

mmirate:
Oh, I see. Check the "Actions" tab, there should be a startup failure listed there because my RTFM'ing failed to yield a correct understanding.

mmirate:
I'll reopen this if I can get it working on my own fork.

mmirate:
A run of this same branch, with my own codecov token, is available at https://github.com/mmirate/tokio-imap/pull/1.

djc:
Wow, that's convoluted! Can you revert the code change in a separate commit? Don't worry about the commit history, I'll squash it all together.

mmirate:
Done. To be clear, are you saying that the misparse that was described in that comment, is expected behavior given the limitations of the IMAP grammar?

djc:
No, merely saying that adding a comment to the tests doesn't belong in a commit for fixing up codecov breaking CI.

mmirate:
Ah, yes, you're right.

PR #94, opened 2020-10-02
Adding a `BodyStructParser`

yageek:
Regarding #92, I propose to add an helper parser to be able to search out specific component from a `BodyStructure`. This will allow for example to retrieve the body section ID of some attachment from the BOPDYSTRUCTURE and perform a FETCH request in a second time to download only this specific element.

PR #111, opened 2021-01-15
Add a `BodyStructParser`, take two

mmirate:
This is #94, once more, with feeling.


djc:
@mmirate thanks for pushing this across the finish line!

PR #115, opened 2021-03-06
Cows! Cows everywhere!

### LINKED ISSUE #114: Store borrowed fields using `Cow`
The current implementation stores `&'a T` for many fields (I'm particularly thinking of [`ResponseCode`](https://docs.rs/imap-proto/0.12/imap_proto/types/enum.ResponseCode.html), but it applies to others). That makes a lot of sense for parsing, but makes it hard to clone the type to keep it beyond the parsing step.

This could be mitigated by replacing every `&'a T` with `Cow<'a, T>` and then adding
```rust
impl<'a> ResponseCode<'a> {
   pub fn into_owned(self) -> ResponseCode<'static> { }
}
```
to each such type, which just maps all the internal `Cow`'s into their owned variants.

That would make these types a lot more ergonomic to use in more contexts, and would avoid forcing downstream libraries to keep their own owned versions of these types!

jonhoo:
Fixes #114.

### DIFF
--- a/imap-proto/src/types.rs
+++ b/imap-proto/src/types.rs
@@ -121,6 +121,10 @@ impl<'a> ResponseCode<'a> {
             ResponseCode::AppendUid(a, b) => ResponseCode::AppendUid(a, b),
             ResponseCode::CopyUid(a, b, c) => ResponseCode::CopyUid(a, b, c),
             ResponseCode::UidNotSticky => ResponseCode::UidNotSticky,
+            ResponseCode::MetadataLongEntries(v) => ResponseCode::MetadataLongEntries(v),
+            ResponseCode::MetadataMaxSize(v) => ResponseCode::MetadataMaxSize(v),
+            ResponseCode::MetadataTooMany => ResponseCode::MetadataTooMany,
+            ResponseCode::MetadataNoPrivate => ResponseCode::MetadataNoPrivate,
         }
     }
 }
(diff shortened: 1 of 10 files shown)


imap-proto/src/types.rs
```diff
@@ -88,6 +89,37 @@ impl From<u32> for UidSetMember {
...
 }
 
+impl<'a> ResponseCode<'a> {
+    pub fn into_owned(self) -> ResponseCode<'static> {
+        match self {
+            ResponseCode::Alert => ResponseCode::Alert,
+            ResponseCode::BadCharset(v) => ResponseCode::BadCharset(v.map(|vs| {
+                vs.into_iter()
+                    .map(Cow::into_owned)
+                    .map(Cow::Owned)
```
djc: Surely we we can do better than `map().map()`? Maybe `map(|v| Cow::into_owned(v).into())`?

jonhoo: I actually found the former a little nicer to read since it doesn't require the closure, but will change :+1:

jonhoo: Added a helper function — much nicer now!

jonhoo:
:tada: Do you have an estimate for when the next (breaking) release might happen? I'd like to include this change in the upcoming `imap 3.0.0` release if possible.

djc:
I expect that to happen about 5 seconds ago. 👍

PR #116, opened 2021-03-06
Add RFC 5464 response codes

### LINKED ISSUE #113: Missing ResponseCode: LONGENTRIES from METADATA extension
The METADATA IMAP extension defines a LONGENTRIES response code that should probably be added to `ResponseCode`: https://tools.ietf.org/html/rfc5464#section-4.2.1

jonhoo:
Fixes #113.


imap-proto/src/types.rs
```diff
@@ -71,6 +71,10 @@ pub enum ResponseCode<'a> {
     AppendUid(u32, Vec<UidSetMember>),
     CopyUid(u32, Vec<UidSetMember>, Vec<UidSetMember>),
     UidNotSticky,
+    MetadataLongEntries(u64), // RFC 5464, section 4.2.1
+    MetadataMaxSize(u64),     // RFC 5464, section 4.3
+    MetadataTooMany,          // RFC 5464, section 4.3
+    MetadataNoPrivate,        // RFC 5464, section 4.3
```
djc, line 77: Any thoughts on whether this ought to be a single variant that contains another enum value? Perhaps that would make it easier to handle these as a group if desired?

imap-proto/src/parser/rfc3501/mod.rs
```diff
@@ -176,6 +176,10 @@ fn resp_text_code(i: &[u8]) -> IResult<&[u8], ResponseCode> {
             rfc4315::resp_text_code_append_uid,
             rfc4315::resp_text_code_copy_uid,
             rfc4315::resp_text_code_uid_not_sticky,
+            rfc5464::resp_text_code_metadata_longentries,
```
djc: Also let's do `long_entries` etc.

imap-proto/src/types.rs
jonhoo: I don't _think_ they should be split out. They're not generally matched together, and if they were, it'd be _either_ the first _or_ the last three. And it seems awkward to have an enum that covers just the last three, or one that covers all four when only three of them will be matched as a group..

djc:
Thanks!

jonhoo:
No, thank you!

Issue #117, opened 2021-03-11
[Suggestion] RFC 5256 - SORT command


soywod:
It could be nice to parse the `SORT` command as defined in the [RFC-5256](https://tools.ietf.org/html/rfc5256#section-3). I can propose a PR!

djc:
Happy to review your PR!

soywod:
Let me know when a new version is available, so I can use it in my tool!

djc:
I noticed you have some other features on the himalaya roadmap, do any of them require further imap-proto changes? If so, I'd prefer to batch them all up in a single release.

soywod:
The potential changes I see are about this same rfc5256 (the thread section). I don't plan to implement it for now (not even for the v1), so I would say I have nothing more!

djc:
I've pushed out imap-proto 0.13.1.

PR #118, opened 2021-03-11
RFC 5256 - SORT command

soywod:
#117

### DIFF
--- a/imap-proto/src/parser/rfc5256.rs
+++ b/imap-proto/src/parser/rfc5256.rs
@@ -16,15 +16,19 @@ use crate::{parser::core::number, types::MailboxDatum};
 
 /// BASE.7.2.SORT. SORT Response
 ///
-///    Data:       zero or more numbers
+/// Data:       zero or more numbers
 ///
-///       The SORT response occurs as a result of a SORT or UID SORT
-///       command.  The number(s) refer to those messages that match the
-///       search criteria.  For SORT, these are message sequence numbers;
-///       for UID SORT, these are unique identifiers.  Each number is
-///       delimited by a space.
+/// The SORT response occurs as a result of a SORT or UID SORT
+/// command.  The number(s) refer to those messages that match the
+/// search criteria.  For SORT, these are message sequence numbers;
+/// for UID SORT, these are unique identifiers.  Each number is
+/// delimited by a space.
 ///
-///    Example:    S: * SORT 2 3 6
+/// Example:
+///
+/// ```ignore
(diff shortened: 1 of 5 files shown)


imap-proto/src/parser/rfc5256.rs
```diff
@@ -0,0 +1,27 @@
...
+    sequence::{preceded, terminated},
+    IResult,
+};
+
+use crate::{parser::core::number, types::MailboxDatum};
+
+pub(crate) fn mailbox_data_sort(i: &[u8]) -> IResult<&[u8], MailboxDatum> {
+    map(
+        // Technically, trailing whitespace is not allowed here, but multiple
+        // email servers in the wild seem to have it anyway (see #34, #108).
```
djc: Have you actually seen this for this particular extension?

imap-proto/src/parser/rfc5256.rs
```diff
@@ -0,0 +1,27 @@
...
+    bytes::streaming::{tag, tag_no_case},
+    combinator::{map, opt},
+    multi::many0,
+    sequence::{preceded, terminated},
+    IResult,
+};
+
+use crate::{parser::core::number, types::MailboxDatum};
+
+pub(crate) fn mailbox_data_sort(i: &[u8]) -> IResult<&[u8], MailboxDatum> {
```
djc, line 34: Can you add a docstring here with the syntax from the relevant RFC, and maybe a little context on what this is/how it's used? (There are many existing parsers that have stuff like this, see #116 for some examples.)

imap-proto/src/parser/rfc5256.rs
soywod: In fact the `SORT` command is very close to the `SEARCH` command (it kind of extends it). So I guess they share the same "issues". Should I remove?

djc: I guess it's fine to keep, but may make it explicit that this is conjecture based on the similarity to `SEARCH`?

djc:
Thanks!

soywod:
Thank YOU for this great tool!

jonhoo:
@djc Ah, this change was not backwards-compatible, since `MailboxDatum` isn't declared as `#[non_exhaustive]` (even though it probably should be). May be worth adding that attribute and going `0.14`?

djc:
Ah, sorry about that! Will do

djc:
Published 0.14.0 and yanked 0.13.1.

PR #120, opened 2021-04-11
Add into_owned() for Response and MailboxDatum.

mordak:
This is similar to #115 and adds `into_owned()` for `Response` and `MailboxDatum`.


djc:
Thanks!

mordak:
Do you know when this will make it into a release? Thank you!

djc:
I just pushed out 0.14.1. Hope that helps!

mordak:
> I just pushed out 0.14.1. Hope that helps!

It does very much! Thank you! :)

PR #1, opened 2017-11-10
Refresh README

sanmai-NL:
Add more badges. These help marketing the project. Also add a link to the Tokio stack overview page, to help novices.

PR #122, opened 2021-04-13
Fix #1

stavteklassi:
Fixes #1


dario23:
this looks very fishy, @djc you may want to look into this (and cancel the CI jobs if possible?). weird code side-loaded in the last line, "hidden" by base64 encoding.. i already reported the user (has a number of other "Fix #1" PRs to other repos, nothing else), maybe more people want to do that..

PR #124, opened 2021-05-29
Support for GETQUOTA response rfc2087

Simon-Laux:
### Idea / why?
I want that DeltaChat can warn me with a device message when my mailbox is running full.
I asked @link2xt about how I could accomplish this and the answer was that I need to implement parsing of the results into `imap-proto` first and then implement a function for it in `async-imap` before I can implement what I want into `deltachat-core`.

### To Do:
- [X] QUOTA Response
- [x] Fix QUOTAROOT Response parsing tests (help wanted, I'm unable to fix this on my own right-now, nom complains about incomplete input, testing via `crate::parser::rfc3501::response_data` on the other hand works, only testing the standalone function makes problems)

_As I don't really understand completely what I need to do / am doing here, please review it carefully._

### DIFF
--- a/imap-proto/src/parser/rfc2087.rs
+++ b/imap-proto/src/parser/rfc2087.rs
@@ -197,13 +197,13 @@ mod tests {
         );
     }
 
-    fn terminated_qouta_root(i: &[u8]) -> IResult<&[u8], Response> {
+    fn terminated_quota_root(i: &[u8]) -> IResult<&[u8], Response> {
         nom::sequence::terminated(quota_root, nom::bytes::streaming::tag("\r\n"))(i)
     }
 
     #[test]
     fn test_quota_root_without_root_names() {
-        assert_matches! (terminated_qouta_root(b"QUOTAROOT comp.mail.mime\r\n") ,
+        assert_matches! (terminated_quota_root(b"QUOTAROOT comp.mail.mime\r\n") ,
             Ok((_, r)) => {
                 assert_eq!(
                     r,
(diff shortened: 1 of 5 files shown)


imap-proto/src/parser/rfc2087.rs
```diff
@@ -0,0 +1,273 @@
...
+                    })
+                );
+            }
+        );
+    }
+
+    #[test]
+    fn test_quota_root2() {
+        assert_matches! (
+            terminated_qouta_root(b"QUOTAROOT INBOX HU\r\n"),
```
link2xt: `qouta` -> `quota`

imap-proto/src/parser/rfc2087.rs
```diff
@@ -0,0 +1,273 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use assert_matches::assert_matches;
+    use std::borrow::Cow;
+
+    #[test]
+    fn test_quota() {
+        assert_matches! (quota(b"QUOTA \"\" (STORAGE 10 512)") ,
```
djc: Nit: here and elsewhere, please eliminate the space after `assert_matches!`, align the first argument on the next line and eliminate the space before the comma.

djc (commented):
This looks good, thanks! I have one style nit (unfortunately rustfmt doesn't always know how to deal with macros), please fix that up and I'll merge this.

Simon-Laux:
after merging a new version of this crate would be nice/helpful.

djc:
I've published 0.14.2. Note that I've removed the new `Capability::Quota` variant again, since it seems this can be handled just fine with the existing `Capability::Atom` variant, and this avoids breaking API compatibility.

PR #126, opened 2021-06-01
add derive clone to quota structs/types

Simon-Laux:
I think that's the last thing I needed to do in `imap-proto` for get-quota support.

PS: a new 1.14.3 release of imap-proto including this would be awesome :innocent:  :tada:


djc:
Ah, sorry I forgot to merge this!

djc:
Published 0.14.3.

PR #131, opened 2021-07-29
allow status responses to contain no response-text

frankstolle:
As already mentioned in the comments of method resp_text there is the possibility to receive such a response without any text. In that case, it makes sense that the response has no trailing space. 

The Apple iCloud IMAP-servers for example sends a `* BYE\r\n` on logout.


djc:
Thanks!

PR #134, opened 2021-10-23
Add name attributes

MichaelMcDonnell:
Any programs that use the name attributes from the LIST command
currently have to match on strings. This change set adds a
`NameAttribute` enumeration and parses the name attributes so you can
match on the enumeration instead.


imap-proto/src/parser/rfc3501/name_attributes.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+fn extension(i: &[u8]) -> IResult<&[u8], NameAttribute> {
+    map(extension_str, |s| {
+        NameAttribute::Extension(Cow::Borrowed(s))
+    })(i)
+}
+
+fn name_attribute(i: &[u8]) -> IResult<&[u8], NameAttribute> {
+    alt((no_inferiors, no_select, marked, unmarked, extension))(i)
```
djc: I think we can trivially inline the alternative parsers here, and that would have my preference. It should significantly cut down on the size of this module, and then I'd prefer to just inline the contents of this module into `rfc3501/mod.rs`, above the `mailbox_list()` definition.

imap-proto/src/types.rs
```diff
@@ -711,6 +714,48 @@ impl<'a> BodyExtMPart<'a> {
     }
 }
 
+/// The name attributes are returned as part of a LIST response described in
+/// [RFC 3501 section 7.2.2](https://tools.ietf.org/html/rfc3501#section-7.2.2).
+#[derive(Debug, Eq, PartialEq, Clone)]
```
djc, line 719: Should we mark this non-exhaustive? Are you aware of any extensions that extend this?

djc (commented):
This is a nice improvement, thanks! The implementation looks pretty good, modulo some minor style suggestions.

imap-proto/src/parser/rfc3501/name_attributes.rs
MichaelMcDonnell: Ok I'll have a go at inlining the alternative parsers and moving the code. Thanks!

imap-proto/src/types.rs
MichaelMcDonnell: Good point! Yes, I think we should mark it non-exhaustive since there are extensions the name attributes. I would like to add support for RFC 6154 Special-Use Mailboxes which will add more enum values (e.g. to identify the junk/spam folder).

It brings up an interesting problem by the way. When support for an extension is later added through a new enum value, then code using the string in the `Extension` value will no longer be run. The string `Extension` value will be consumed during parsing and will instead show up as a new enum value. I have unfortunately not been able to figure out a good solution to that.

djc: I think there's not much we can do about it in the end. IIRC there is a way to make sure downstream matches do match all variants, so that would at least notify users there are new variants to match on.

imap-proto/src/parser/rfc3501/name_attributes.rs
MichaelMcDonnell: I moved and inlined the code. Not sure if I went overboard with the inlining :smile:

PR #135, opened 2021-10-29
Add special-use mailboxes name attributes

MichaelMcDonnell:
This change only adds the name attributes from RFC 6154 Special-Use Mailboxes and not any of the other features.

It should now be possible in a client app to identify the spam folder without matching on the name of the mailbox. Matching on the name of the mailbox is not reliable because it might be localized to the user's language.

### DIFF
--- a/imap-proto/src/parser/rfc3501/mod.rs
+++ b/imap-proto/src/parser/rfc3501/mod.rs
@@ -245,10 +245,12 @@ fn mailbox_data_exists(i: &[u8]) -> IResult<&[u8], MailboxDatum> {
 
 fn name_attribute(i: &[u8]) -> IResult<&[u8], NameAttribute> {
     alt((
+        // RFC 3501
         value(NameAttribute::NoInferiors, tag_no_case(b"\\Noinferiors")),
         value(NameAttribute::NoSelect, tag_no_case(b"\\Noselect")),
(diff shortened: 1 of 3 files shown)


imap-proto/src/parser/rfc3501/mod.rs
```diff
@@ -249,6 +249,13 @@ fn name_attribute(i: &[u8]) -> IResult<&[u8], NameAttribute> {
         value(NameAttribute::NoSelect, tag_no_case(b"\\Noselect")),
         value(NameAttribute::Marked, tag_no_case(b"\\Marked")),
         value(NameAttribute::Unmarked, tag_no_case(b"\\Unmarked")),
+        value(NameAttribute::All, tag_no_case(b"\\All")),
```
djc, line 254: Here and in the other instances where all the variants are enumerated, would you mind adding a line comment before each of the RFC 3501, this PR and misc blocks?

MichaelMcDonnell: I have added line comments to separate the name attributes into blocks. That should make it a little easier to tell where the name attributes came from. Is that what you wanted?

djc:
Exactly right, thanks!

MichaelMcDonnell:
Great! Thanks!

PR #136, opened 2021-11-02
Restore test coverage from `NameAttribute` change

MichaelMcDonnell:
My recent changes to add `NameAttribute` parsing lowered the test coverage. I have added a simple test to make sure `into_owned` returns the same value but just with different ownership. That should restore the test coverage to what it was before.


djc:
Thanks!

MichaelMcDonnell:
Wow, that was a fast merge! Thanks!

PR #137, opened 2021-12-26
Add support for googles x-gm-labels extension

estriv:
Hi,

I was using your crate to interact with the Google IMAP server. As I heavily rely on [Google's IMAP extensions](https://developers.google.com/gmail/imap/imap-extensions) I started implementing them with X-GM-LABELS.

I would be glad if you accept the changes. Further implementation will be done as I find time and the need.

PR #138, opened 2021-12-27
Add support for googles x-gm-labels extension

estriv:
Followup of https://github.com/djc/tokio-imap/pull/137 with `cargo fmt` commit.

### DIFF
--- a/imap-proto/src/parser/gmail.rs
+++ b/imap-proto/src/parser/gmail.rs
@@ -1,0 +1,56 @@
+use std::borrow::Cow;
+
+use nom::IResult;
+use nom::branch::alt;
+use nom::bytes::streaming::tag_no_case;
+use nom::sequence::preceded;
+use nom::combinator::map;
+
+use crate::{AttributeValue, MailboxDatum};
+
+use super::core::{parenthesized_list, quoted_utf8};
+use super::rfc3501::flag;
+
+pub(crate) fn x_gm_label_list(i: &[u8]) -> IResult<&[u8], Vec<Cow<str>>> {
+  preceded(
+      tag_no_case("X-GM-LABELS "),
+      parenthesized_list(map(alt((flag, quoted_utf8)), Cow::Borrowed)),
+  )(i)
+}
+
+pub(crate) fn msg_att_x_gm_labels(i: &[u8]) -> IResult<&[u8], AttributeValue> {
+  map(
+    x_gm_label_list,
+    AttributeValue::XGmLabels
+  )(i)
+}
+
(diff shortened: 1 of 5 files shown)


imap-proto/src/parser/rfc3501/mod.rs
```diff
@@ -62,14 +64,14 @@ fn mailbox(i: &[u8]) -> IResult<&[u8], &str> {
     })(i)
 }
 
-fn flag_extension(i: &[u8]) -> IResult<&[u8], &str> {
+pub(crate) fn flag_extension(i: &[u8]) -> IResult<&[u8], &str> {
```
djc: Increasing the visibility here does not seem necessary?

imap-proto/src/parser/gmail.rs
```diff
@@ -0,0 +1,52 @@
...
+use nom::combinator::map;
+use nom::sequence::preceded;
+use nom::IResult;
+
+use crate::{AttributeValue, MailboxDatum};
+
+use super::core::{parenthesized_list, quoted_utf8};
+use super::rfc3501::flag;
+
+pub(crate) fn x_gm_label_list(i: &[u8]) -> IResult<&[u8], Vec<Cow<str>>> {
```
djc: Rather than using the `x_gm_` prefix on these functions, let's just replace that with `gmail`, since that seems the intended meaning.

imap-proto/src/types.rs
```diff
@@ -293,6 +297,7 @@ pub enum Attribute {
     Rfc822Size,
     Rfc822Text,
     Uid,
+    XGmLabels,
```
djc: Should have a comment linking to https://developers.google.com/gmail/imap/imap-extensions#access_to_gmail_labels_x-gm-labels.

imap-proto/src/types.rs
```diff
@@ -327,6 +332,7 @@ pub enum AttributeValue<'a> {
     Rfc822Size(u32),
     Rfc822Text(Option<Cow<'a, [u8]>>),
     Uid(u32),
+    XGmLabels(Vec<Cow<'a, str>>),
```
djc: This should also have a comment linking to the extension "specification".

djc (commented):
This mostly looks fine, some minor nits on how to do extensions like this.

@link2xt I'm happy to discuss how we could better support extensions. I'm inclined to accept support for GMail extensions because GMail is so widely used; these extensions are probably getting used quite a bit compared to some extensions that are actually specified in RFCs.

imap-proto/src/types.rs
estriv: I propose naming the enum `GmailLabels` following https://github.com/djc/tokio-imap/pull/138#discussion_r775630683.

djc:
Thanks!

PR #142, opened 2022-04-23
Add support for RFC 4314/2086 ACL extension to imap-proto

urkle:
Implement parsing of the ACL response types specified in RFC 4314/2086 (and the Annotation right from RFC 5257)

ACL response
LISTRIGHTS response
MYRIGHTS response

### DIFF
--- a/imap-proto/src/parser/rfc4314.rs
+++ b/imap-proto/src/parser/rfc4314.rs
@@ -1,0 +1,119 @@
+//!
+//! Current
+//! https://tools.ietf.org/html/rfc4314
+//!
+//! Original
+//! https://tools.ietf.org/html/rfc2086
+//!
+//! The IMAP ACL Extension
+//!
+
+use std::borrow::Cow;
+
+use nom::{
+    bytes::streaming::tag_no_case,
+    character::complete::{space0, space1},
+    combinator::map,
+    multi::separated_list0,
+    sequence::{preceded, separated_pair, tuple},
+    IResult,
+};
+
+use crate::parser::core::astring_utf8;
+use crate::parser::rfc3501::mailbox;
+use crate::types::*;
+
+/// 3.6. ACL Response
+/// ```ignore
+/// acl_response  ::= "ACL" SP mailbox SP acl_list
+/// ```
+pub(crate) fn acl(i: &[u8]) -> IResult<&[u8], Response> {
+    let (rest, (_, _, mailbox, acls)) = tuple((
+        tag_no_case("ACL"),
+        space1,
+        map(mailbox, Cow::Borrowed),
+        acl_list,
+    ))(i)?;
+
+    Ok((rest, Response::Acl(Acl { mailbox, acls })))
+}
+
+/// ```ignore
+/// acl_list  ::= *(SP acl_entry)
+/// ```
+fn acl_list(i: &[u8]) -> IResult<&[u8], Vec<AclEntry>> {
+    preceded(space0, separated_list0(space1, acl_entry))(i)
+}
+
+/// ```ignore
+/// acl_entry ::= SP identifier SP rights
+/// ```
+fn acl_entry(i: &[u8]) -> IResult<&[u8], AclEntry> {
+    let (rest, (identifier, rights)) = separated_pair(
+        map(astring_utf8, Cow::Borrowed),
+        space1,
+        map(astring_utf8, map_text_to_rights),
+    )(i)?;
+
+    Ok((rest, AclEntry { identifier, rights }))
+}
+
+/// 3.7. LISTRIGHTS Response
+/// ```ignore
+/// list_rights_response  ::= "LISTRIGHTS" SP mailbox SP identifier SP required_rights *(SP optional_rights)
+/// ```
+pub(crate) fn list_rights(i: &[u8]) -> IResult<&[u8], Response> {
+    let (rest, (_, _, mailbox, _, identifier, _, required, optional)) = tuple((
+        tag_no_case("LISTRIGHTS"),
+        space1,
+        map(mailbox, Cow::Borrowed),
+        space1,
+        map(astring_utf8, Cow::Borrowed),
+        space1,
+        map(astring_utf8, map_text_to_rights),
+        list_rights_optional,
+    ))(i)?;
+
+    Ok((
+        rest,
+        Response::ListRights(ListRights {
+            mailbox,
+            identifier,
+            required,
+            optional,
+        }),
+    ))
+}
+
+fn list_rights_optional(i: &[u8]) -> IResult<&[u8], Vec<AclRight>> {
+    let (rest, items) = preceded(space0, separated_list0(space1, astring_utf8))(i)?;
+
+    Ok((
+        rest,
+        items
+            .into_iter()
+            .flat_map(|s| s.chars().map(|c| c.into()))
+            .collect(),
+    ))
+}
+
+/// 3.7. MYRIGHTS Response
+/// ```ignore
+/// my_rights_response  ::= "MYRIGHTS" SP mailbox SP rights
+/// ```
+pub(crate) fn my_rights(i: &[u8]) -> IResult<&[u8], Response> {
+    let (rest, (_, _, mailbox, _, rights)) = tuple((
+        tag_no_case("MYRIGHTS"),
+        space1,
+        map(mailbox, Cow::Borrowed),
+        space1,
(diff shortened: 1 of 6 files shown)


imap-proto/src/parser/rfc4314.rs
```diff
@@ -0,0 +1,109 @@
...
+//! Current
+//! https://tools.ietf.org/html/rfc4314
+//!
+//! Original
+//! https://tools.ietf.org/html/rfc2086
+//!
+//! The IMAP ACL Extension
+//!
+
+use nom::{
```
djc, line 13: Please keep the std dependency in a separate block of imports, before the block containing the nom imports.

imap-proto/src/parser/rfc4314.rs
```diff
@@ -0,0 +1,109 @@
...
+    )(i)?;
+
+    Ok((rest, ACLEntry { identifier, rights }))
+}
+
+/// 3.7. LISTRIGHTS Response
+/// ```ignore
+/// listrights_response  ::= "LISTRIGHTS" SP mailbox SP identifier SP required_rights *(SP optional_rights)
+/// ```
+pub(crate) fn listrights(i: &[u8]) -> IResult<&[u8], Response> {
```
djc: I think this name should be spelled `list_rights` in Rust (I see you already spelled it `ListRights` in a variant name, so that would be more consistent). Please apply that to all the instances, including test names.

imap-proto/src/types.rs
```diff
@@ -45,6 +45,9 @@ pub enum Response<'a> {
     Quota(Quota<'a>),
     QuotaRoot(QuotaRoot<'a>),
     Id(Option<HashMap<Cow<'a, str>, Cow<'a, str>>>),
+    ACL(ACL<'a>),
```
djc: `ACL` should be spelled `Acl` in all of the relevant types and variants (as suggested in https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case).

imap-proto/src/types.rs
```diff
@@ -925,6 +931,9 @@ impl<'a> QuotaRoot<'a> {
     }
 }
 
+pub mod acls;
```
djc: Style nit: I like to keep these closer to the top, in a separate block below the imports.

imap-proto/src/parser/tests.rs
```diff
@@ -49,6 +49,336 @@ fn test_name_attributes() {
...
+                }
+            )
+        }
+        rsp => panic!("unexpected response {:?}", rsp),
+    }
+}
+
+/// Test the MYRIGHTS response from RFC 4314/2086
+#[test]
+fn test_acl_myrights_response() {
```
djc: Similar to `list_rights`, Rust identifiers should use `my_rights`.

imap-proto/src/parser/rfc4314.rs
```diff
@@ -0,0 +1,109 @@
...
+        space1,
+        map(mailbox, Cow::Borrowed),
+        space1,
+        map(astring_utf8, map_text_to_rights),
+    ))(i)?;
+
+    Ok((rest, Response::MyRights(MyRights { mailbox, rights })))
+}
+
+fn map_text_to_rights(i: &str) -> ACLRightList {
```
djc: This should be a method or `From`/`Into` impl on `AclRightList`, instead.

imap-proto/src/types/acls.rs
```diff
@@ -0,0 +1,259 @@
...
+            ACLRight::Annotation => 'n',
+            ACLRight::OldCreate => 'c',
+            ACLRight::OldDelete => 'd',
+            ACLRight::Custom(c) => c,
+        }
+    }
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct ACLRightList {
```
djc: It's not obvious that we should have this type inside imap-proto. I think we should instead use a `Vec<AclRight>` which directly reflects what the server sent us.

djc (commented):
Thanks for working on this! Some feedback below, most of it is stylistic.

imap-proto/src/types.rs
urkle: fully agree with you there.   This was just due to refactoring while I was building it out as I originally had all the types in this file, but it was getting too large.

imap-proto/src/parser/rfc4314.rs
urkle: I already have the From defined for it. (after extracting it out).  I replaced the uses of this with direct calls to the into() instead.

imap-proto/src/types/acls.rs
urkle: Ok.. I factored this out

imap-proto/src/parser/rfc4314.rs
urkle: ok.. so with removing of the ACLRightList struct this method is back again as one cannot create a "From" or "Into" implementation between a &str and a Vec<> :(

urkle:
@djc thanks for the review.. I've updated the branch with your requested changes.

imap-proto/src/parser/rfc4314.rs
```diff
@@ -0,0 +1,118 @@
...
+        }),
+    ))
+}
+
+fn list_rights_optional(i: &[u8]) -> IResult<&[u8], AclRightList> {
+    let (rest, items) = preceded(space0, separated_list0(space1, astring_utf8))(i)?;
+
+    let v: String = items.into_iter().flat_map(|s| s.chars()).collect();
+
+    Ok((rest, map_text_to_rights(&v)))
```
djc: This seems to have an intermediate allocation for no good reason. Can you rewrite to something that converts directly?

imap-proto/src/types/acls.rs
```diff
@@ -0,0 +1,179 @@
...
+            AclRight::Administer => 'a',
+            AclRight::Annotation => 'n',
+            AclRight::OldCreate => 'c',
+            AclRight::OldDelete => 'd',
+            AclRight::Custom(c) => c,
+        }
+    }
+}
+
+pub(crate) type AclRightList = Vec<AclRight>;
```
djc: I'd prefer to avoid the alias; IMO these usually make it harder to understand the code.

imap-proto/src/parser/rfc4314.rs
```diff
@@ -0,0 +1,118 @@
...
+        map(astring_utf8, map_text_to_rights),
+    ))(i)?;
+
+    Ok((rest, Response::MyRights(MyRights { mailbox, rights })))
+}
+
+/// helper routine to map a string to a vec of AclRights
+fn map_text_to_rights(i: &str) -> Vec<AclRight> {
+    i.chars()
+        .into_iter()
```
djc: I think the type returned by `chars()` is an `Iterator`, so no need to call `into_iter()` on it?

djc:
Also I thought I wrote a comment about this, but please order the type definitions in the new module as top down. Most complex type (with dependencies on other types) first, simplest type (`AclRight`) last.

imap-proto/src/types/acls.rs
urkle: yeah.. sorry that was supposed to be removed..   I have it removed now.

imap-proto/src/parser/rfc4314.rs
urkle: yeah.. let me see if I can rework this.

urkle: ok.. I reworked it to remove the String and to have it convert more directly.

imap-proto/src/parser/rfc4314.rs
```diff
@@ -0,0 +1,120 @@
...
+    ))
+}
+
+fn list_rights_optional(i: &[u8]) -> IResult<&[u8], Vec<AclRight>> {
+    let (rest, items) = preceded(space0, separated_list0(space1, astring_utf8))(i)?;
+
+    Ok((
+        rest,
+        items.into_iter()
+            .flat_map(|s| map_text_to_rights(s))
```
djc: We need to resolve this lint still. I think the fix here is not to use `map_text_to_rights()` but to just use `s.chars().map(|c| c.into())` directly, which would avoid allocating a `Vec` which will be dropped again immediately after.

djc (commented):
Looking good, almost there! Just need to fix up the lints.

djc:
0.16.1 is on crates.io now!

PR #144, opened 2022-12-12
Accept char ']' in flag list

### LINKED ISSUE #220: Unable to parse status response
Some of [himalaya](https://github.com/soywod/himalaya/issues/269) users reported me this error while trying to display their messages. The error occurs there:

```rust
let fetches = self
    .sess()?
    .fetch(&range, "(ENVELOPE FLAGS INTERNALDATE)")
    .context(format!(r#"cannot fetch messages within range "{}""#, range))?;
```

It looks like the response from the IMAP server cannot be parsed properly. The problem is that:

1. I cannot reproduce their error (it looks like it fails in a specific context)
2. I cannot see which response fails because it is not included in the final error message:
https://github.com/jonhoo/rust-imap/blob/6808dfef79e26945366794dc16e66cc0ad3d2a15/src/error.rs#L229

Any idea?

soywod:
Fixes https://github.com/jonhoo/rust-imap/issues/220.

### DIFF
--- a/imap-proto/src/parser/rfc3501/mod.rs
+++ b/imap-proto/src/parser/rfc3501/mod.rs
@@ -72,7 +72,17 @@ fn flag_extension(i: &[u8]) -> IResult<&[u8], &str> {
 }
 
 pub(crate) fn flag(i: &[u8]) -> IResult<&[u8], &str> {
-    alt((flag_extension, atom))(i)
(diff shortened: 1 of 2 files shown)


soywod:
I guess it is ready for review, let me know!

djc:
Thanks!

PR #148, opened 2023-06-23
Add support for google x-gm-msgid extension

lancel0:
Add support for the X-GM-MSGID extension

Gmail provides a unique message ID for each email so that a unique message may be identified across multiple folders.

I made changes similar to https://github.com/djc/tokio-imap/pull/138 and https://github.com/jonhoo/rust-imap/pull/225 then tested that msg ids were correctly fetched from google.


djc (approved):
LGTM, thanks!

PR #152, opened 2023-11-06
fix: allow trailing SP at the end of responses

link2xt:
Outlook server sends STATUS responses with a space at the end
even though RFC 3501 formal syntax does not allow it.


djc:
See #153.

PR #159, opened 2024-01-27
Fix parsing of NIL metadata values

link2xt:
map_opt always fails if second argument returns None.


djc (approved):

djc:
Thanks!

djc:
BTW I'm open to PRs that make the parser recoverable where possible.

djc:
IIRC it's probably possible to impose some meta structure that just looks for `\r\n` and knows how to work around `DATA` blocks so it can recover to the next `\r\n`.

PR #168, opened 2024-05-01
fix: allow whitespace at the end of ID response list

link2xt:
https://archiveopteryx.org/ returns it in the `ID` response.


djc:
Thanks!

PR #170, opened 2024-05-04
Bump imap-proto version to 0.16.5

link2xt:


djc (approved):

djc:
* Published imap-proto v0.16.5 at registry `crates-io`
* [new tag]         imap-proto-0.16.5 -> imap-proto-0.16.5

PR #176, opened 2025-07-04
removed check for null characters in literal function

AdSkipper1337:
Hi - I am developing an app using rust-imap to get emails.

I implemented a full mailbox sync for my clients mailbox and started running it, after a few thousand mails in I got a crash. I started investigating and I tracked down the culprit all the way to these lines. 
If I remove them it works.

What the lines did was that they throw an error if there is a null chracter somewhere in the email.

Now let me explain: What my client has in his mailbox is a corrupted mail that contains a binary within it. 
That cetainly does not meet any standars - however, in practice its bound to be encountered. 
My AST parsers and sanitizers that are run after that handle the mess just fine and delete it as is propper, leaving the base message readable, just like it is readable in Thunderbird and webmailers. 
Without the checker it works just fine, and it is an unnecessary crash  that does not reflect real world data.

The comment said: 
// FIXME: what ErrorKind should this have?
Probably this shouldn't be an error at all.
I am not sure if you are encountering this issue in production, and if this handles genuine cases for you.
But just silent failure is definetly not expected behavior so probably this is pretty rare.

I thought about sanitizing the string in this case but it would hurt performance too much to either make it mutable or allocate a new array here.

So here is my working fix, with food for thought. 

Best!

### DIFF
--- a/imap-proto/src/parser/core.rs
+++ b/imap-proto/src/parser/core.rs
@@ -98,20 +98,7 @@ pub fn literal(input: &[u8]) -> IResult<&[u8], &[u8]> {
 
     let (remaining, data) = take(count)(remaining)?;
 
-    if !data.iter().all(|byte| is_char8(*byte)) {
-        // FIXME: what ErrorKind should this have?
-        return Err(nom::Err::Error(nom::error::Error::new(
-            remaining,
-            nom::error::ErrorKind::Verify,
-        )));
-    }
-
     Ok((remaining, data))
-}
-
-/// CHAR8 = %x01-ff ; any OCTET except NUL, %x00
-pub fn is_char8(i: u8) -> bool {
-    i != 0
 }
 
 // ----- astring ----- atom (roughly) or string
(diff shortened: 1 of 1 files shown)


djc (commented):
Suggest you filter out null characters (which I think are never allowed in IMAP connections?) before hitting the parser.

AdSkipper1337:
Thanks for the quick response!

First: I am well aware that this is not a situation that meets the standard. But the question is how to handle it IF it happens, I can not change the fact that someone sends a mail that is messed up in this way and that the IMAP connection is sending it; Also webmailers and thunderbird support that mail anyway so I fear I will have to as well. If its something that shouldn't be given anyway then returning something with null characters would be a sane response.

I just thought about it, and played arround with it - but I think that won't actually work because there could be an attachment in the mail which is a binary file containing valid null characters. And - as far as I understand it - rust-imap uses your library to extract it, so if I put a filter in readline then it will destroy the attachments. 
So I would have to parse it in order to know where to filter and where not to filter.
I have started investigating this since yesterday evening so I am not an expert on either of these libraries, so please correct me if I am wrong here.

djc (approved):
Okay, I'll merge this if you can additionally provide a test.

AdSkipper1337:
Done. Thank you very much!

djc:
Thanks!

Issue #178, opened 2025-08-09
Issue when parsing select response flags from surgemail server


cry-inc:
Hello, I created an IMAP client app that uses the `imap-proto` crate via `async-imap`. A user [reported an IMAP issue](https://github.com/cry-inc/dmarc-report-viewer/issues/51) when connecting to his server running surgemail. I was able to reproduce the issue with a [minimal IMAP client example](https://github.com/chatmail/async-imap/issues/123) on my own test instance of surgemail version 80e.

The issue boils down to surgemail sending an extra space at the end of the flags list before the closing `)`:
```
< SELECT "INBOX"
<
> * 1 EXISTS
> * 0 RECENT
> * OK [UNSEEN 1] first unseen message
> * OK [UIDVALIDITY 1754171061] Uid epoch
> * OK [UIDNEXT 2] Predicted next uid
> * FLAGS (\Answered \Flagged \Deleted \Draft \Seen $Forwarded )
> * OK [PERMANENTFLAGS (\Answered \Flagged \Deleted \Draft \Seen $Forwarded )] Limited
> A0002 OK [READ-WRITE] SELECT completed
```
[link2xt](https://github.com/link2xt) (maintainer of `async-imap`) [documented the issue](https://github.com/modern-email/defects/issues/48) and sent me here.

I was able to trigger the parser error in question using this small test:
```
#[test]
fn test_flags() {
    super::response_data(b"* FLAGS (\\Answered \\Flagged \\Deleted \\Draft \\Seen $Forwarded )\r\n").unwrap();
}
```
Is this a case where you would consider adding a workaround for the surgemail bug in `imap-proto`?

djc:
Sure, send me a PR with a fix that includes your test and I'll review. Make sure to add some comments to explain why.

PR #179, opened 2025-08-10
Workaround for surgemail bug with spaces at the end of flag list

cry-inc:
Added workaround for surgemail bug with unwanted spaces at the end of flag lists reported in #178

### DIFF
--- a/imap-proto/src/parser/core.rs
+++ b/imap-proto/src/parser/core.rs
@@ -2,9 +2,9 @@ use nom::{
     branch::alt,
     bytes::streaming::{escaped, tag, tag_no_case, take, take_while, take_while1},
     character::streaming::{char, digit1, one_of},
-    combinator::{map, map_res},
+    combinator::{map, map_res, opt},
     multi::{separated_list0, separated_list1},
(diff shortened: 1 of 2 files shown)


imap-proto/src/parser/core.rs
```diff
@@ -212,7 +212,14 @@ where
     F: FnMut(&'a [u8]) -> IResult<&'a [u8], O, E>,
     E: nom::error::ParseError<&'a [u8]>,
 {
-    delimited(char('('), separated_list0(char(' '), f), char(')'))
+    delimited(
+        char('('),
+        separated_list0(char(' '), f),
+        preceded(
+            many0(char(' ')), // Surgemail sometimes sends a space before the closing bracket.
```
djc: Can we do `opt` instead of `many0`?

djc (commented):
This mostly makes sense to me, thanks.

imap-proto/src/parser/core.rs
cry-inc: Yes, that makes more sense. Done!

djc (approved):
Thanks!

PR #188, opened 2025-12-28
Small dependency optimizations

paolobarbolini:


djc (approved):
Thanks!

I don't think anyone's actually using tokio-imap...

Issue #190, opened 2026-04-20
Parsing failure: Strict whitespace handling on FETCH responses (extra space before closing parenthesis)


rustmailer:
I’m currently using imap-proto for a mail archiving project and have encountered a parsing issue with FETCH responses. It seems the parser is strictly rejecting responses that contain a space before the closing parenthesis in certain fields.

```
#[test]
fn test_fetch2222() {
    parse_response(b"* 11784 FETCH (UID 87567 INTERNALDATE \"06-May-2023 18:48:30 +0200\" RFC822.SIZE 5799845 )\r\n").unwrap();
}
```
The parser fails on this pattern:
`RFC822.SIZE 5799845 )`

If I remove the space `(RFC822.SIZE 5799845)`, it parses correctly.
```
#[test]
fn test_fetch2222() {
    parse_response(b"* 11784 FETCH (UID 87567 INTERNALDATE \"06-May-2023 18:48:30 +0200\" RFC822.SIZE 5799845)\r\n").unwrap();
}
```

It appears the grammar definition is strictly expecting the closing ) immediately after the value. However, we have observed that some real-world IMAP servers include this space.

Is this strictness intentional per the RFC? If so, would you be open to a PR that makes the parser more tolerant of optional whitespace in these response structures to improve compatibility with real-world servers?

Thanks for your time and for maintaining this library!

djc:
It is intentional but I'm open to targeted lenience. Please submit a PR including a comment.

rustmailer:
@djc 
I’ve submitted a small PR to address the whitespace issue. The change is very targeted—it simply introduces a minor lenience for trailing whitespace before the closing parenthesis.

I’ve verified that all existing unit tests pass, so this change should not have any negative impact on other parsing logic. When you have a moment, could you please take a look?

Thanks again for your guidance!

duesee:
Just stumbled upon this and wanted to invite to add this defect to #modern-email/defects, too. We maintain a collection of defects there to make implementing real-world protocols easier.

djc:
Fixed in

- #191

PR #191, opened 2026-04-20
fix(parser): allow trailing whitespace before closing parenthesis #190

rustmailer:
Some real-world IMAP servers include extra whitespace before the closing parenthesis. This change makes the parser more robust by allowing zero or more spaces.

### DIFF
--- a/imap-proto/src/parser/core.rs
+++ b/imap-proto/src/parser/core.rs
@@ -1,7 +1,7 @@
 use nom::{
     branch::alt,
     bytes::streaming::{escaped, tag, tag_no_case, take, take_while, take_while1},
-    character::streaming::{char, digit1, one_of},
+    character::streaming::{char, digit1, one_of, space0},
     combinator::{map, map_res, opt},
     multi::{separated_list0, separated_list1},
     sequence::{delimited, preceded, tuple},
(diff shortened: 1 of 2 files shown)


imap-proto/src/parser/core.rs
```diff
@@ -204,7 +204,13 @@ where
     F: FnMut(&'a [u8]) -> IResult<&'a [u8], O, E>,
     E: nom::error::ParseError<&'a [u8]>,
 {
-    delimited(char('('), separated_list1(char(' '), f), char(')'))
+    delimited(
+        char('('),
+        separated_list1(char(' '), f),
+        // Targeted lenience: Some real-world IMAP servers
```
djc, line 210: Can you be more concrete about which server? If feasible, did you file an issue against it?

djc (approved):

imap-proto/src/parser/core.rs
rustmailer: imap.free.fr

rustmailer: No, this comes from feedback from a user.

> If feasible, did you file an issue against it?

djc: Apparently this might be some (older?) version of Cyrus IMAP.

djc:
If you want a release for this, can you submit a PR with a version bump?

PR #193, opened 2026-04-28
imap-proto: handle unknown FETCH attributes (RFC 8474 EMAILID/THREADID)

x-hannibal:
## Problem

RFC-compliant IMAP servers can include extension attributes in `FETCH` responses that `imap-proto` does not yet explicitly support. Known cases:

- **RFC 8474 §5 OBJECTID**: `EMAILID`, `THREADID` — sent by Apache James 3.9, Dovecot (with OBJECTID plugin), and others
- **RFC 8514 §2**: `SAVEDATE`
- Any future extension attribute

When one of these attributes appears in a `FETCH` list, `msg_att`'s `alt()` exhausts all known parsers. Because nom 7's `alt` returns the **last** alternative's error (not the deepest), the caller sees a `TakeWhile1` error from `response_tagged`'s `imap_tag` — which tries to match `*` as a tag character and fails. The entire response is dropped.

## Fix

Add `msg_att_unknown` as the **last** arm of `msg_att`'s `alt()`. It consumes:

- a single-level parenthesised value `(...)` — covers `EMAILID (Mxxx)`, `THREADID (Txxx)`
- an nstring — covers `THREADID NIL`, `SAVEDATE "date-time"`
- a bare atom or number — fallback for any remaining scalar value

and returns `AttributeValue::Unknown` so the rest of the `FETCH` attribute list can be parsed without error.

`AttributeValue::Unknown` is added to the enum as a documented variant. The name and raw value are consumed and discarded; callers that need the actual value for a specific extension should open a dedicated tracking issue or PR to add first-class support.

## Tests

Four new tests in `parser/tests.rs`:

| Test | Input |
|------|-------|
| `test_fetch_rfc8474_emailid` | `EMAILID (Mxxx)` alongside `UID` + `BODY[]<0>` |
| `test_fetch_rfc8474_threadid` | `THREADID (Txxx)` alongside `UID` + `BODY[]<0>` |
| `test_fetch_rfc8474_threadid_nil` | `THREADID NIL` (RFC 8474 §5.2 allows NIL) |
| `test_fetch_rfc8474_emailid_and_threadid` | both in the same `FETCH` response |

All 86 existing tests continue to pass.

## Notes

- `msg_att_unknown` is intentionally the **last** alternative so it never shadows an explicitly supported attribute.
- Single-level parentheses only — deeply nested values (which do not appear in any known extension attribute grammar) are not handled. A dedicated parser can be added per RFC if needed.

### DIFF
--- a/imap-proto/src/parser/rfc3501/mod.rs
+++ b/imap-proto/src/parser/rfc3501/mod.rs
@@ -566,16 +566,53 @@ fn msg_att_uid(i: &[u8]) -> IResult<&[u8], AttributeValue<'_>> {
 //                   "UID" SP uniqueid
 //                     ; MUST NOT change for a message
 
+// RFC 8474 §5.1 — EMAILID
+//   "EMAILID" SP "(" objectid ")"
+// objectid = 1*ASTRING-CHAR (RFC 8474 §3).  EMAILID is non-optional: a
+// compliant server MUST be able to provide one for any stored message,
+// so the wire form `EMAILID NIL` is not allowed by the RFC.  If a
+// non-compliant server emits it anyway, this parser fails and the
+// catch-all `msg_att_unknown` below absorbs the attribute.
+fn msg_att_emailid(i: &[u8]) -> IResult<&[u8], AttributeValue<'_>> {
+    map(
+        preceded(
+            tag_no_case("EMAILID "),
+            paren_delimited(map_res(take_while1(is_astring_char), from_utf8)),
+        ),
+        |id| AttributeValue::EmailId(Cow::Borrowed(id)),
+    )(i)
+}
+
+// RFC 8474 §5.2 — THREADID
+//   "THREADID" SP ( "(" objectid ")" / nil )
+// THREADID may be NIL, which the RFC mandates for messages that do not
+// currently have a thread association.  We map NIL → `None`.
+fn msg_att_threadid(i: &[u8]) -> IResult<&[u8], AttributeValue<'_>> {
+    map(
+        preceded(
+            tag_no_case("THREADID "),
+            alt((
+                map(nil, |_| None),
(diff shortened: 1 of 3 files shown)


imap-proto/src/parser/rfc3501/mod.rs
```diff
@@ -565,6 +565,41 @@ fn msg_att_uid(i: &[u8]) -> IResult<&[u8], AttributeValue<'_>> {
 //                   "BODY" section ["<" number ">"] SP nstring /
 //                   "UID" SP uniqueid
 //                     ; MUST NOT change for a message
+
+// Catch-all for RFC extension attributes not explicitly handled above.
+//
+// RFC-compliant servers (Apache James, Stalwart, Dovecot) may include
+// attributes from extensions that post-date this crate, for example:
+//   - RFC 8474 §5 OBJECTID: EMAILID / THREADID
```
djc: Can you just add support for these two directly in addition to the unknown support?

x-hannibal: Done in `3b26248`. Added `AttributeValue::EmailId(Cow<'a, str>)` and `AttributeValue::ThreadId(Option<Cow<'a, str>>)` with dedicated parsers (`msg_att_emailid`, `msg_att_threadid`) ordered before `msg_att_unknown` in the `alt()`. Asymmetric API — EMAILID is non-optional per RFC 8474 §5.1, ThreadId carries `Option` for the §5.2 NIL form. The four existing RFC 8474 tests now assert on the typed variants; added `test_fetch_unknown_savedate_fallback` so the `Unknown` path stays covered.

djc:
Thanks!

x-hannibal:
Happy to contribute — thanks for the review!

PR #194, opened 2026-05-03
fix(parser): tolerate 8-bit bytes in strings and literals

rudolfschmidt:
## Summary

Real-world IMAP servers (Dovecot, Cyrus) regularly emit 8-bit bytes in BODYSTRUCTURE responses — typically MIME filenames in non-UTF-8 encodings like Latin-1 or ISO-8859-9 (e.g. Turkish `Görüntü1.png` from Outlook). The current strict 7-bit ASCII parser rejects these, breaking the parse of a single bad mail and desyncing the entire bulk fetch stream — taking down all subsequent FETCH responses in the same buffer.

Two changes:

1. **`is_char` relaxed to accept any non-NUL byte** instead of `0x01..=0x7F`. Matches real-world server behavior; RFC 6855 / IMAP4rev2 (RFC 9051) formally allow UTF-8 in these positions.

2. **`*_utf8` functions return `Cow<'_, str>` with lossy decoding** instead of `&str` with strict `from_utf8`. Affected: `string_utf8`, `nstring_utf8`, `astring_utf8`, `quoted_utf8`, `text`, plus `mailbox`, `resp_text`, `trailing_resp_text`, `mailbox_list`, `id_param`, `id_param_list_not_nil`, `id_param_list`. Callers that previously wrapped `&str` results in `Cow::Borrowed` have been simplified.

**Note**: this is a public-API change (`&str` → `Cow<'_, str>` in the `*_utf8` functions). I think it's the right tradeoff — strict UTF-8 enforcement at the parser level is too strict for real-world IMAP — but happy to discuss alternatives (e.g. parallel `*_lossy` variants, or `unsafe { from_utf8_unchecked }` to keep `&str` API).

## Reproduction

A BODYSTRUCTURE with an 8-bit filename literal:

```
* 15 FETCH (BODYSTRUCTURE ("image" "png" ("name" {8}\r\n<G><F6><r><FC><n><t><FC>1) ...
```

Where the 8 bytes after `{8}\r\n` are `0x47 0xF6 0x72 0xFC 0x6E 0x74 0xFC 0x31` (Turkish `Görüntü1` in Latin-1, valid IMAP literal but not valid UTF-8).

On master this fails with `code: TakeWhile1` because:
- `quoted` parser uses `take_while1(is_text_char)` which rejects bytes `>= 0x80`
- Even when `quoted` succeeds, `from_utf8` in `string_utf8` fails on Latin-1 sequences

With the fix the parser proceeds and the filename is decoded lossily (replacement chars for invalid UTF-8 sequences).

## Test plan

- [x] All 82 existing `imap-proto` tests pass
- [x] New regression test `test_body_structure_with_8bit_literal_filename` passes with the fix
- [x] Same test reverted-to-master state fails with the exact `TakeWhile1` error from the field bug report

Run:

```
cargo test -p imap-proto
```

### DIFF
--- a/imap-proto/src/parser/core.rs
+++ b/imap-proto/src/parser/core.rs
@@ -8,6 +8,7 @@ use nom::{
     IResult,
 };
 
+use std::borrow::Cow;
 use std::str::{from_utf8, FromStr};
 
 // ----- number -----
@@ -61,9 +62,17 @@ pub fn string(i: &[u8]) -> IResult<&[u8], &[u8]> {
     alt((quoted, literal))(i)
 }
 
-// string bytes as utf8
-pub fn string_utf8(i: &[u8]) -> IResult<&[u8], &str> {
-    map_res(string, from_utf8)(i)
+#[inline]
+fn lossy_str(bytes: &[u8]) -> Cow<'_, str> {
+    String::from_utf8_lossy(bytes)
+}
+
+// string bytes as utf8 — falls back to lossy decoding when the literal
+// or quoted contents include non-UTF-8 bytes (e.g. Latin-1 / ISO-8859-9
+// MIME filenames in BODYSTRUCTURE responses). Strict from_utf8 would
+// fail, breaking the entire response stream on a single bad mail.
+pub fn string_utf8(i: &[u8]) -> IResult<&[u8], Cow<'_, str>> {
+    map(string, lossy_str)(i)
 }
 
 // quoted = DQUOTE *QUOTED-CHAR DQUOTE
@@ -79,9 +88,9 @@ pub fn quoted(i: &[u8]) -> IResult<&[u8], &[u8]> {
     )(i)
 }
 
-// quoted bytes as utf8
-pub fn quoted_utf8(i: &[u8]) -> IResult<&[u8], &str> {
-    map_res(quoted, from_utf8)(i)
+// quoted bytes as utf8 — lossy, see comment on string_utf8.
+pub fn quoted_utf8(i: &[u8]) -> IResult<&[u8], Cow<'_, str>> {
+    map(quoted, lossy_str)(i)
 }
 
 // quoted-specials = DQUOTE / "\"
@@ -108,9 +117,9 @@ pub fn astring(i: &[u8]) -> IResult<&[u8], &[u8]> {
     alt((take_while1(is_astring_char), string))(i)
 }
 
-// astring bytes as utf8
-pub fn astring_utf8(i: &[u8]) -> IResult<&[u8], &str> {
-    map_res(astring, from_utf8)(i)
+// astring bytes as utf8 — lossy, see comment on string_utf8.
+pub fn astring_utf8(i: &[u8]) -> IResult<&[u8], Cow<'_, str>> {
+    map(astring, lossy_str)(i)
 }
 
 // ASTRING-CHAR = ATOM-CHAR / resp-specials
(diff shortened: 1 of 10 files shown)


djc:
I'm generally happy to improve compatibility particularly if it aligns with RFC 9051. I want to be certain that relaxing a low-level functions doesn't affect callers that should be more strict according to RFC 9051, though. Have you reviewed callers to make sure relaxing is compliant for all of them?

rudolfschmidt:
Yes, I walked the callers. The `*_utf8` functions are used by:

- BODYSTRUCTURE field parsers (filenames, content-types, content-IDs, languages, params, MD5) — RFC 9051 §7.5.2 explicitly allows UTF-8 here; legacy non-UTF-8-aware servers sometimes emit Latin-1 / ISO-8859.
- `mailbox` + `mailbox_list` → RFC 9051 §5.1 mandates UTF-8 mailbox names.
- `resp_text` / `id_param*` / X-GM-LABELS → also user-visible strings.

For every caller the question is "what to do with bytes the server sent that aren't valid UTF-8". The previous behavior (parse-failure) cascades because a single bad BODYSTRUCTURE desyncs the entire FETCH response stream. Lossy decode at the parser level is strictly an improvement: clients that need stricter validation can re-validate downstream from the `Cow<str>`.

I'm not aware of any caller in the tree that benefits from strict rejection over lossy decode + downstream validation. Happy to add a `*_strict` parallel variant if you'd prefer to keep the strict path as an opt-in for callers that genuinely need it.

imap-proto/src/parser/core.rs
```diff
@@ -61,9 +62,17 @@ pub fn string(i: &[u8]) -> IResult<&[u8], &[u8]> {
...
+// string bytes as utf8 — falls back to lossy decoding when the literal
+// or quoted contents include non-UTF-8 bytes (e.g. Latin-1 / ISO-8859-9
+// MIME filenames in BODYSTRUCTURE responses). Strict from_utf8 would
+// fail, breaking the entire response stream on a single bad mail.
+pub fn string_utf8(i: &[u8]) -> IResult<&[u8], Cow<'_, str>> {
+    map(string, lossy_str)(i)
+}
+
+#[inline]
+fn lossy_str(bytes: &[u8]) -> Cow<'_, str> {
```
djc: Nit: it looks like prevailing style is bottom-up (caller after callee), so please move `lossy_str()` above `string_utf8()`.

imap-proto/src/parser/core.rs
```diff
@@ -164,22 +173,26 @@ pub fn nil(i: &[u8]) -> IResult<&[u8], &[u8]> {
...
 // TEXT-CHAR = <any CHAR except CR and LF>
 pub fn is_text_char(c: u8) -> bool {
     is_char(c) && c != b'\r' && c != b'\n'
 }
 
-// CHAR = %x01-7F
-//          ; any 7-bit US-ASCII character,
-//          ;  excluding NUL
-// From RFC5234
+// CHAR per RFC5234 is 7-bit US-ASCII excluding NUL, but real-world IMAP
```
djc: Nit: please retain the grammar from the spec, and add your further context/explanation after.

djc (approved):
Please squash your commits, and add a commit with a semver-incompatbile version bump.

rudolfschmidt:
Squashed and added the version bump commit. Also addressed both nits — moved `lossy_str` above `string_utf8` and restored the original RFC 5234 grammar block on `is_char` with the explanation appended after.

The bump commit also updates `tokio-imap/Cargo.toml`'s `imap-proto` dep from `0.16` to `0.17` so the workspace resolves.

rudolfschmidt:
Anything left blocking the merge? Happy to do additional changes.

djc:
Thanks! I'll do some tweaks and publish the release some time next week.

PR #198, opened 2026-07-18
Add UIDONLY and MESSAGELIMIT response parsing

gabeosx:
Yahoo publicly documents two primary mailbox-access modes. Its default Limited mode exposes only a partial view for folders above `MESSAGELIMIT`; UID mode, activated with `ENABLE UIDONLY`, allows clients to fetch and operate on all messages using UIDs. Correct UIDONLY and MESSAGELIMIT support is therefore necessary for `imap-proto` consumers that need complete large-mailbox exports from Yahoo rather than a limited view.

RFC 9586 defines `UIDFETCH` responses whose leading number is a UID, not a message sequence number. Representing these responses as ordinary `FETCH` responses would make that distinction ambiguous for clients. RFC 9738 defines the `MESSAGELIMIT` response code used when a server limits an operation.

This PR adds parser support for `UIDFETCH`, `UIDREQUIRED`, and `MESSAGELIMIT`. Existing `parse_response` handling of `ENABLED` responses remains unchanged. A separate opt-in `parse_response_with_enabled` entry point lets callers distinguish `ENABLED` responses from `CAPABILITY` responses when needed.

This is the parser layer required by such clients; it does not add client commands or export logic to `tokio-imap`.

Tests cover ENABLED compatibility, UIDFETCH metadata and literals, malformed and incomplete input, nonzero UID requirements, numeric boundaries, UIDREQUIRED, and tagged and untagged MESSAGELIMIT response codes.

Checks:

- `cargo test --all-features --offline`
- `cargo fmt --all -- --check`
- `cargo clippy --all-targets --offline -- -D warnings`
- `cargo check --manifest-path imap-proto/Cargo.toml --lib --all-features --offline`
- `git diff --check`

References:

- Yahoo IMAP documentation: https://senders.yahooinc.com/developer/documentation/
- RFC 5161: https://www.rfc-editor.org/rfc/rfc5161.html
- RFC 9586: https://www.rfc-editor.org/rfc/rfc9586.html
- RFC 9738: https://www.rfc-editor.org/rfc/rfc9738.html


gabeosx:
hi @djc was wondering if you had a chance to review. I'm trying to integrate UIDONLY and MESSAGELIMIT support with the intent to contribute to some downstream projects allowing Yahoo mail backup to work correctly. ty

djc:
Unfortunately imap-proto isn't much of a priority for me right now -- unless you have funding to help support the project it'll just take me a while to get to reviewing this.

PR #206, opened 2026-09-15
Add a test for LOGIN response with an en dash in it

link2xt:
This is a test for a problem reported for Delta Chat using imap-proto 0.16.5. The issue appears to be already fixed in commit e599fd4acd30a06f8df9027798d528fe73d7a737 (pull request https://github.com/djc/imap-proto/pull/194) but is not part of any release because imap-proto 0.17.0 is not released yet and the latest version 0.16.7 does not contain the fix.

The test is superfluous at this point, but I think it might be useful anyway to avoid regressions. Original fix changed many places in the code,
but only added two tests test_resp_text_lossy_decode_8bit and  test_body_structure_with_8bit_literal_filename and the problem there was with a filename, while in case of mail.systemausfall.org the problem happens during parsing the LOGIN response already.

The same response 
`* OK [CAPABILITY IMAP4rev1 LOGIN-REFERRALS ID ENABLE IDLE SASL-IR LITERAL+ AUTH=PLAIN AUTH=LOGIN AUTH=XOAUTH2] Logged in - go ahead!` is returned even before LOGIN as a banner,
you can receive it by running "openssl s_client -connect mail.systemausfall.org:993". ID command says it is a Dovecot server, just with the banner apparently configured manually:
```
? ID ()
* ID ("name" "Dovecot")
```


link2xt:
@djc I have also tested that current `main` has the problem fixed by upgrading to it in https://github.com/chatmail/async-imap/pull/144, if you could tag a new release of imap-proto that would fix the problem https://github.com/chatmail/async-imap/issues/145 for us downstream.

djc (approved):

djc:
Will try to release it soon.

